PR #1, opened 2021-09-09
Updated translations

BudsieBuds:
Hi,

I've updated the English translation, mostly following Microsoft Terminology because this is a Windows-only application. The goal here is to, hopefully, be more consistent and be more in line with native Windows apps. Also updated the Dutch translation along the way. This one is based on the new English file.

Maybe you should consider running an instance of Weblate for easier translation and attracting more translators. It supports the rc file type being used here.

Hope this is useful, thanks for an awesome piece of software!


assarbad:
Hi, thanks for the effort. I'll see how I can first inspect your changes (GitHub won't show them) and then merge this. I don't actually use Git (and refuse to do so for most of my sparetime projects), but this should not be your concern.

**Thanks for the pointer to Weblate.** This is in fact a service (and software) which I hadn't heard off. Looks good and would indeed lower the bar for contributors looking to improve the translations. Also once I have the software in that shape it would ease things for me. I'll take a closer look to see how I can set this up on our own domain.

assarbad:
Oh, one more thing @BudsieBuds. I was wondering if you found any updated glossary lists from MS. The ones I got from MSDN some years ago don't seem to have any updates. I have a VS Pro subscription at my disposal, so if you have a keyword for the updated ones, I can also search there.

Thanks.

BudsieBuds:
@assarbad Sorry, I totally forgot to respond. You're welcome, hope it's useful!

You can find the terminology on the [Microsoft Language Portal](https://www.microsoft.com/en-us/language). There's an online translate tool, and downloads like terminology in tbx-format, UI translations etc. You'll probably find what you're looking for there.

On a side note, I wonder why you refuse to use Git?

PR #2, opened 2023-01-27
Add DisplayVersion to uninstaller registry hive

frehov:
Minor convenience fix for users of winget.

Sets the String `DisplayVersion` to the current full version in the uninstaller registry hive located at `HKCU\Software\Microsoft\Windows\CurrentVersion\Uninstall\WinDirStat`

Reason for change.
Winget can currently install windirstat, but loses track immediately of the version and relies upon the version being available on the uninstaller entry under the key `DisplayVersion`

Change has been tested locally by adding the registry key manually and verifying that winget now lists the version when listing installed apps. Change has been verified by setting a lower version (1.1.1) and verifying that winget upgrade includes windirstat 1.1.2 as an available upgrade.

image

### DIFF
--- a/setup/WDS.nsi
+++ b/setup/WDS.nsi
@@ -267,6 +267,7 @@ Function CreateUninstallEntry
   WriteRegExpandStr HKCU "Software\Microsoft\Windows\CurrentVersion\Uninstall\WinDirStat" "InstallLocation" "$INSTDIR"
   WriteRegStr HKCU "Software\Microsoft\Windows\CurrentVersion\Uninstall\WinDirStat" "DisplayName" "${sVersionFull}"
   WriteRegStr HKCU "Software\Microsoft\Windows\CurrentVersion\Uninstall\WinDirStat" "DisplayIcon" "$INSTDIR\windirstat.exe,0"
+  WriteRegStr HKCU "Software\Microsoft\Windows\CurrentVersion\Uninstall\WinDirStat" "DisplayVersion" "${dVersionMajor}.${dVersionMinor}.${dVersionRev}"
   WriteRegDWORD HKCU "Software\Microsoft\Windows\CurrentVersion\Uninstall\WinDirStat" "dwVersionMajor" "${dVersionMajor}"
   WriteRegDWORD HKCU "Software\Microsoft\Windows\CurrentVersion\Uninstall\WinDirStat" "dwVersionMinor" "${dVersionMinor}"
   WriteRegDWORD HKCU "Software\Microsoft\Windows\CurrentVersion\Uninstall\WinDirStat" "dwVersionRev" "${dVersionRev}"


setup/WDS.nsi
```diff
@@ -267,6 +267,7 @@ Function CreateUninstallEntry
   WriteRegExpandStr HKCU "Software\Microsoft\Windows\CurrentVersion\Uninstall\WinDirStat" "InstallLocation" "$INSTDIR"
   WriteRegStr HKCU "Software\Microsoft\Windows\CurrentVersion\Uninstall\WinDirStat" "DisplayName" "${sVersionFull}"
   WriteRegStr HKCU "Software\Microsoft\Windows\CurrentVersion\Uninstall\WinDirStat" "DisplayIcon" "$INSTDIR\windirstat.exe,0"
+  WriteRegStr HKCU "Software\Microsoft\Windows\CurrentVersion\Uninstall\WinDirStat" "DisplayVersion" "${dVersionMajor}.${dVersionMinor}.${dVersionRev}"
   WriteRegDWORD HKCU "Software\Microsoft\Windows\CurrentVersion\Uninstall\WinDirStat" "dwVersionMajor" "${dVersionMajor}"
   WriteRegDWORD HKCU "Software\Microsoft\Windows\CurrentVersion\Uninstall\WinDirStat" "dwVersionMinor" "${dVersionMinor}"
   WriteRegDWORD HKCU "Software\Microsoft\Windows\CurrentVersion\Uninstall\WinDirStat" "dwVersionRev" "${dVersionRev}"
```
Bertaz, line 273: Should also include VersionMajor and VersionMinor:
```nsi
+ WriteRegDWORD HKCU "Software\Microsoft\Windows\CurrentVersion\Uninstall\WinDirStat" "VersionMajor" "${dVersionMajor}"
+ WriteRegDWORD HKCU "Software\Microsoft\Windows\CurrentVersion\Uninstall\WinDirStat" "VersionMinor" "${dVersionMinor}"
```
<sup>See [microsoft](https://learn.microsoft.com/en-us/windows/win32/msi/uninstall-registry-key), [nsis](https://nsis.sourceforge.io/Docs/AppendixD.html#useful_add_uninst_infos), [comment](https://github.com/microsoft/winget-cli/issues/1255#issuecomment-902873661).</sup>

setup/WDS.nsi
```diff
@@ -267,6 +267,7 @@ Function CreateUninstallEntry
   WriteRegExpandStr HKCU "Software\Microsoft\Windows\CurrentVersion\Uninstall\WinDirStat" "InstallLocation" "$INSTDIR"
   WriteRegStr HKCU "Software\Microsoft\Windows\CurrentVersion\Uninstall\WinDirStat" "DisplayName" "${sVersionFull}"
   WriteRegStr HKCU "Software\Microsoft\Windows\CurrentVersion\Uninstall\WinDirStat" "DisplayIcon" "$INSTDIR\windirstat.exe,0"
+  WriteRegStr HKCU "Software\Microsoft\Windows\CurrentVersion\Uninstall\WinDirStat" "DisplayVersion" "${dVersionMajor}.${dVersionMinor}.${dVersionRev}"
```
Bertaz, line 270: Would be better to use version already formatted as a string, probably `${sVersionS}` or either of `${sVersion_Ansi}` and `${sVersion_Unicode}` if they always point to the same version

Bertaz (changes requested):

setup/WDS.nsi
Bertaz: Can't find other instances of `dwVersionMajor`, `dwVersionMinor`, `dwVersionRev` and `dwVersionBuild` anywhere, they don't appear to be valid for the uninstall reg.
May be easier to remove dw from `dwVersionMajor` and `dwVersionMinor` (only those two, because `VersionRev` and `VersionBuild` aren't valid)
Unless those dw… keys are used somewhere else in the code. I couldn't find it.

frehov: Maybe those dw keys are from a this kind of struct?
https://learn.microsoft.com/en-us/windows/win32/api/winnt/ns-winnt-osversioninfoa

But I can fix the VersionMajor and VersionMinor parts.

setup/WDS.nsi
frehov: Nice, didn't know there was a preformatted version string.
As long as it's consistent, it shouldn't really matter I think.

But can change to `${sVersionS}`, I didn't find any references where it's being set, so I opted for the safe route of constructing it manually

Just to be sure I downloaded from fosshub just now, and the filename comes out as `windirstat1_1_2_setup.exe` but maybe the most important part is that the what is injected to the properties of the installer is the full version i.e `<major>.<minor>.<revision/patch>.<build>`

setup/WDS.nsi
assarbad: Nope, `OSVERSIONINFO` is an entirely different beast, although I'll concede that the naming was probably inspired by it a bit 😉

assarbad:
Hmm, using `winget` myself (and encountering the same issue with a number of other software, too).

That said, I see no issue in getting this change in. However:

1. it doesn't help any current users and will only take effect upon next upgrade
2. the current in-progress work uses WiX and therefore MSI (also in the hope that a Microsoft tool will be more inclined to work properly with installers using their technology ...)

Microsoft has changed the rules for what values should be written how and where so many times, it's no fun.

But this makes sense, I'll see how I can get this in, despite using Hg.

assarbad:
@frehov do you want to upgrade your PR? Otherwise I'll pull it and then add the other two values myself.

But as I pointed out, it will only really take effect for the next upgrade 😑

frehov:
Sorry, life's been a bit busy.
If I haven't updated the PR within the next 24 hours, feel free to just pull it in.
I _should_ be able to fire off a quick commit tomorrow at work.

Appreviate the feedback though :)
Stoked for any updates that may come down the line.

setup/WDS.nsi
Bertaz: > Just to be sure I downloaded from fosshub just now, and the filename comes out as `windirstat1_1_2_setup.exe` but maybe the most important part is that the what is injected to the properties of the installer is the full version i.e `<major>.<minor>.<revision/patch>.<build>`

The `1_1_2` string comes from `${sVersionFile}`, see:
https://github.com/windirstat/windirstat/blob/351490639619ee2273a18d05105a1f11fe627341/setup/WDS.nsi#L19
The FileVersion in the properties of the installer (1.1.2.79) comes from `${sVersionS}`, see:
https://github.com/windirstat/windirstat/blob/351490639619ee2273a18d05105a1f11fe627341/setup/WDS.nsi#L86
This for some reason is different from the version in the program's exe (1.1.2.80).


> But can change to `${sVersionS}`, I didn't find any references where it's being set, so I opted for the safe route of constructing it manually

Yeah, those are probably set internally, I agree that without knowing constructing it manually was the best choice.
Maybe @assarbad can say which one should be preferred `${dVersionMajor}.${dVersionMinor}.${dVersionRev}`, `${sVersionS}` or `${sVersion_Ansi}`/`${sVersion_Unicode}`?

Bertaz:
I'm going to open a new PR including @frehov's commit and my suggestions. You can either merge this one, wait for my merge from upstream, and then merge mine or merge directly mine and close this one, whichever you prefer. Former would be better for proper attribution.

vibcodes:
Hi all, any idea when this will be fixed in the stable install? I am still facing this issue in the latest version in windows

JaneX8:
Can this please be fixed? I still have this issue.

PR #3, opened 2023-02-18
Add DisplayVersion, VersionMajor and VersionMinor to uninstall HKCU

Bertaz:
Copy of PR #2 by @frehov 
Additionally added `VersionMajor` and `VersionMinor`.

Original pr:
> Minor convenience fix for users of winget.
> 
> Sets the String `DisplayVersion` to the current full version in the uninstaller registry hive located at `HKCU\Software\Microsoft\Windows\CurrentVersion\Uninstall\WinDirStat`
> 
> Reason for change. Winget can currently install windirstat, but loses track immediately of the version and relies upon the version being available on the uninstaller entry under the key `DisplayVersion`
> 
> Change has been tested locally by adding the registry key manually and verifying that winget now lists the version when listing installed apps. Change has been verified by setting a lower version (1.1.1) and verifying that winget upgrade includes windirstat 1.1.2 as an available upgrade.
> 
> image


Bertaz:
Draft because I'm waiting on the answer to this https://github.com/windirstat/windirstat/pull/2#discussion_r1110468096
> … which one should be preferred `${dVersionMajor}.${dVersionMinor}.${dVersionRev}`, `${sVersionS}` or `${sVersion_Ansi}`/`${sVersion_Unicode}`?

mi-hol:
I'd suggest:
change to ${sVersionS}, 

> I didn't find any references where it's being set, so I opted for the safe route of constructing it manually

the reasoning "safe route of constructing it manually" is flawed from my opinion because it is just a workaround for "I didn't find any references where it's being set" and this workaround is doing more harm than good in the long run!

PR #10, opened 2024-01-27
partially re-enable checking compressed file size

AartBluestoke:
this PR enables GetCompressedFileSize, but only uses it when the File Attributes say the file could return a different compressed size. 

https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getcompressedfilesizea says GetCompressedFileSize reports different size for compressed or sparse files, so lets check for those flags.

On my laptop there is perhaps a 15% slowdown in  runtime - an example below shows 22 second vs 24 seconds when considering a 1GB difference in 1.6TB/1.3 million files
(most of the time is spent "suspended" according to the bottom bar, and the code without this modification sometimes takes longer than the code without this modification)

Comparison of such a run (main branch top, my PR below - note the updated size is 0.9 GB smaller, accounting for sparse/compressed files.

image

### DIFF
--- a/windirstat/FileFindWDS.cpp
+++ b/windirstat/FileFindWDS.cpp
@@ -42,23 +42,24 @@ DWORD CFileFindWDS::GetAttributes() const
 // Wrapper for file size retrieval
 // This function tries to return compressed file size whenever possible.
 // If the file is not compressed the uncompressed size is being returned.
-ULONGLONG CFileFindWDS::GetCompressedLength() const
+ULONGLONG CFileFindWDS::GetCompressedLength(DWORD FileAttributes) const
 {
-#if 0 // TODO: make this an option (the compressed size instead of "normal" size
-    ULARGE_INTEGER ret;
-    ret.LowPart = ::GetCompressedFileSize(GetFilePath(), &ret.HighPart);
+    // if it is possible for disk-size to differ from reported-size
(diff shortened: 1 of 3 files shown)


AartBluestoke:
test run on :
- windows 11 
- SSD drives 
- onedirve with moderate amount of information
- "large" database files, with "pre-allocate" set (sparse).
- haven't investigated what actually causes the 900mb difference.

AartBluestoke:
I am happy for this code to be used under any licence the main code is released under.

NoMoreFood:
I've reworked this area of the code quite a bit over the last few weeks.  I will take a look at your changes and see if we can integrate these in some sort of optional capacity as to not harm performance.

NoMoreFood:
@AartBluestoke Disregard my previous request.  Try compiling the current source and see if it works for you.  The default is to show the compressed sizes.  You can enable the compressed sizes under the Advanced tab in the configuration options - post an issue in the Issues area if you have any problems.  Thanks for your suggestion and submission!

AartBluestoke:
@NoMoreFood - i've tested the latest code branch - i'll rase some minor issues found, but re: compresson, the new way is so much better - not needing that recursive call. The main branch runs in about 10 seconds on my laptop - twice as fast as it was before (naturally compression makes no difference now)

PR #58, opened 2024-05-15
Update Czech translation for beta build 2024-05-07

PEERSOFTdev:
This translation is based upon the original WinDirStat Czech translation, adding and updating strings and terminology.


NoMoreFood (approved):
Thank you!

PR #92, opened 2024-10-12
Trying to refine the automation by making it conditional

assarbad:
- The run of the `winget-releaser` step is now conditional and should depend on whether or not the respective tag "namespace" matches the intended action (otherwise it should silently skip `winget-releaser`)


assarbad:
@NoMoreFood the test PR from my fork https://github.com/microsoft/winget-pkgs/pull/182334

PR #95, opened 2024-10-16
Correct names of server operating systems

Hermholtz:
I would like to correct server operating system names to accurately reflect the official product naming.


NoMoreFood:
Merged; thanks!

PR #97, opened 2024-10-17
Fix IDS_EXTENSION_MISSING portuguese inconsistency

### LINKED ISSUE #96: Inconsistensy in IDS_EXTENSION_MISSING translation in portuguese
IDS_EXTENSION_MISSING uses "Não Extensão" to translate "No Extension", which is a literal, however incorrect translation.

PedroBittarBarao:
IDS_EXTENSION_MISSING currently uses "Não Extensão" to translate "No Extension", which refers to a file without extension. 
This is incorrect, as it should be "Sem Extensão", just as IDS_POLICY_NOREFRESH uses "Sem atualização" to translate "No Refresh".
Fix #96


NoMoreFood:
Thanks! I used AI for some of these so I appreciate the corrections.  There will probably be a new release in the next week that will include your change.

PR #187, opened 2025-01-12
Patch for pervasive bug found in language translation files

harryytm:
Changed IDS_PAGE_TREELIST_AUTOSIZE to IDS_PAGE_GENERAL_COLAUTOSIZE on line 166 of affected language files and moved the line to 156 to match lexical sort.


NoMoreFood:
Nice find.  Thanks for the fix!

PR #208, opened 2025-02-06
Refactored to show size units in binary prefix

harryytm:

PR #197, opened 2025-01-18
Add New Keyboard Shortcuts

harryytm:
Added some keyboard shortcuts that make sense to me, to make WinDirStat more keyboard friendly, see if that also useful for others.

These shortcut keys are still open for change, please feedback for changing them to something else, or give suggestion to add more keyboard shortcut.

- F10 for Show/Hide Toolbar
- F11 for Show/Hide Statusbar
- Shift+F5 for refresh all
- Alt+Enter for File Properties
- Ctrl+I for Settings
- Ctrl+Alt+E for Run Elevated
- Ctrl+Alt+L for Load Results From CSV
- Ctrl+Alt+S for Save Results to CSV

PR #209, opened 2025-02-06
Add Swedish Language To Resource File and Update Chinese Traditional (Hong Kong)

harryytm:


harryytm:
PRs dependency as follows: #202 -> #209 -> #208 -> #197

PR #216, opened 2025-02-18
UI Finetunes

harryytm:
- Change default columns width of EntensionList
- Change item width calculation count limit to use LargeFileCount and have a range limit of 50 to 10000
- Change AutomaticallyResizeColumns of TreeList to align with auto fit column content (double click column divider)
- Increase GetSubItemWidth width to 3390 which tested enough to display 255 characters long double-byte file name including the drive letter


windirstat/Controls/TreeListControl.cpp
```diff
@@ -698,10 +698,13 @@ void CTreeListControl::ExpandItem(const int i, const bool scroll)
 
         // The calculation of item width is very expensive for
         // very large lists so limit calculation based on the
-        // first few bunch of visible items
-        if (COptions::AutomaticallyResizeColumns && scroll && c < 50)
+        // large file display count which have a range limit
+        // of 50 to 10000, to ensure all items displayed on
+        // the large file list would auto fit the width of
+        // the file path
+        if (COptions::AutomaticallyResizeColumns && scroll && c < max(min(COptions::LargeFileCount,10000),50))
```
NoMoreFood: This would cause GetSubItemWidth() to be called on up to COption::LargeFileCount times on all the inherited controls which can slow down certain operations like expanding very, very large directories.

windirstat/Controls/OwnerDrawnListControl.cpp
```diff
@@ -509,7 +509,7 @@ bool COwnerDrawnListControl::IsShowSelectionAlways() const
 int COwnerDrawnListControl::GetSubItemWidth(COwnerDrawnListItem* item, const int subitem)
 {
     CClientDC dc(this);
-    CRect rc(0, 0, 1000, 1000);
+    CRect rc(0, 0, 3390, 1000);
```
NoMoreFood: Why this specific number?

harryytm: > Why this specific number?

Used the the debugger to pause at TreeListControl.cpp:716, when expending duplicate file tree with 2 files that have 255 double-byte characters long file name file, found out maxwidth was 3389, and roundup it to nearest 10.

Can make it 4000 just to be safe, but not sure if larger number would impact performance or not.
image

windirstat/Controls/TreeListControl.cpp
harryytm: I underhand that would slow down certain operations like expanding very, very large directories, but this is the only solution I found to ensure the auto resize column would set a column width wide enough for all items on the large file list to show the whole path on expending the tree, not just only sampling the first 50 items.

The max valid number of LargeFileCount I found is 10000, I have tried to collapse and expend the Large File list to trigger the loop that call GetSubItemWidth() with LargeFileCount set to 10000 in options, which took me about 5 seconds to display (CPU: i7 4770k), which I think is a good compromise.

Any better solutions to have the same results?

harryytm: On the other hand I found that there is a bug that allow setting LargeFileCount over 10000 on runtime in Options, which is why I put a range limit from 50-10000 to avoid the loop run 99999 times.

I tried without the range limit on TreeListControl.cpp:705  and set LargeFileCount to 99999, would end up 99999 files was populated on large file list, that took me about a minute to expand the large files list tree with auth auto resize columns enabled.

harryytm:
image
Tested 10000 loops of GetSubItemWidth() costs 2291ms on Intel Core i7 4770k.

NoMoreFood:
I will take a better look at this next weekend.  Work has been busy.

Neustradamus:
@NoMoreFood: Have you progressed on it?

windirstat/Pages/PageAdvanced.cpp
```diff
@@ -117,7 +117,7 @@ void CPageAdvanced::OnOK()
     COptions::ExcludeProtectedFile = (FALSE != m_SkipProtectedFile);
     COptions::UseBackupRestore = (FALSE != m_UseBackupRestore);
     COptions::ScanningThreads = m_ScanningThreads + 1;
-    COptions::LargeFileCount = std::stoi(m_LargestFileCount.GetString());
+    COptions::LargeFileCount = min(std::stoi(m_LargestFileCount.GetString()),10000);
```
NoMoreFood: Is the thought here to just provide a reasonable maximum?

windirstat/Controls/TreeListControl.cpp
```diff
@@ -697,11 +697,13 @@ void CTreeListControl::ExpandItem(const int i, const bool scroll)
...
 
         // The calculation of item width is very expensive for
-        // very large lists so limit calculation based on the
-        // first few bunch of visible items
-        if (COptions::AutomaticallyResizeColumns && scroll && c < 50)
+        // very large lists so limit calculation based on 50%
+        // of the LargeFileCount, to ensure all items displayed
+        // on the large file list have the correct column width
+        // to show the entire file path
+        if (COptions::AutomaticallyResizeColumns && scroll && c < max(COptions::LargeFileCount,100) * 0.5)
```
NoMoreFood: Can you test what performance impact this has on expanding a very large directory like C:\Windows\WinSxs?

PR #245, opened 2025-05-08
Create FUNDING.yml

moshemoshe137:


digdilem-work:
Seconded:

Actually ended up here as I'm looking for foss projects to donate to, and struggling to find a way to donate financially to this!

Suggest using Buy me a coffee or Github's FOSS donation tools.

NoMoreFood:
Thanks for the suggestion.  I suppose I can save up so I can meet Oliver in person one day.

PR #247, opened 2025-05-15
Allow -/+ with Ctrl modifier to do Select Parent / Reselect Child

harryytm:
- Allow -/+ with Ctrl modifier to do Select Parent / Reselect Child
- Update Select Parent / Reselect Child keyboard shortcut hints in language files


harryytm:
Thisn shortcut would convenient laptop users with a keyboard without numpad.

PR #249, opened 2025-05-21
Update license.txt

harryytm:
- update license.txt to the current version on gnu.org
- add spaces to some lines to make the title visually center on the about box


NoMoreFood:
Adding the spaces is a bit of a hack.  How does it look if you change the font for that textbox to a monospaced font instead?

harryytm:
> Adding the spaces is a bit of a hack. How does it look if you change the font for that textbox to a monospaced font instead?

Using monospace font would be the most ideal way, but my C++ knowledge is limited.

PR #274, opened 2025-09-03
Enhanced Elevation Prompts & Option For Fast Scanning

harryytm:
- Elevation prompt would
  - Restart in elevation if select Yes
  - Disable UseFastScanEngine setting and untick Fast Scan checkbox if select No
- Prompt user to elevate when running without elevated privileges AND elevation is available in following conditions
   - On startup
   - User enable Fast Scan options in Select Drives
- Prompt user to enable Fast Scan on startup when running with elevated privileges if UseFastScanEngine is disabled in user settings
- Disable the Fast Scan checkbox if elevation is not available
- Add language string of the enable Fast Scan prompt to all supported languages


NoMoreFood:
I do not agree with this "Prompt user to elevate on startup when running without elevated privileges in following conditions".  There are still many advantages to launching elevated even if not using the fast scan option.   The prompt should occur regardless if the user is able to elevate.

harryytm:
> I do not agree with this "Prompt user to elevate on startup when running without elevated privileges in following conditions". There are still many advantages to launching elevated even if not using the fast scan option. The prompt should occur regardless if the user is able to elevate.

@NoMoreFood Corrected the comment to
- Prompt user to elevate when running without elevated privileges in following conditions
   - On startup if UseFastScanEngine is enabled or null in user settings
   - User enable Fast Scan options in Select Drives

The logic is to honor and remember the user choice of choosing NO in the elevlation prompt to reduce annoynace during startup, unless the user choose to enable(tick) FastScan option again in Select Drive, then the prompt would ask the user to elevate instead of only available on startup.

If the user choose Yes in the elevation prompt, it will set UseFastScan setting to true and save to registry/ini and then restart in elevated, the user don't need to manually tick Fast Scan option in Select Drives.
image
I have also added another prompt to remind user to enable Fast Scan when running in elevated on startup, it would apply to the rescan parameter passed  by "Run Elevated" if user selected Yes, without the need of the user go back to Select Drives to enable Fast Scan and restart the scan.
image

harryytm:
@NoMoreFood Thank you for your feedback, I do love to have more input form you to make this modification more refine and polish to improve the user experience, and hopefully it would merge into the main branch.

Let me share my thought on this mods, although personally I would always use WinDirStat in admin right by default, the elevation prompt is not my concern on daily use, but for a normal user that is not technical is a different story, I do think they deserve to have a smarter elevation prompt to well inform them to make the best choice, and would configure the setting according to their wish without extra steps, and do much more than only do elevation and nothing else, to make it more easy and simple to use.

I have extensively tested the user experience to make the controls more streamline and interactive, I am quite happy with the current modifications, but I am still refining the logic of the two prompts to make the prompts appear at the right time, and change the setting for the user automatically to make the UI more convenient and logical.

Hence I currently change the pull to draft, please test the pull request and give feedback to let me know your idea of how to refine it and more align with your vision for the project, please feel free to give your feedback as detail and direct as possible.

harryytm:
I will be working in the supermarket in the next 4 days, but feel free to leave comments in the code, I would try to follow up with my spare time if I still have the energy after work.

harryytm:
@NoMoreFood I have updated the commit message to more accurately reflect the changes. Please feel free to feedback on what need to clarify or change. Thanks.

PR #279, opened 2025-09-10
Fix LargeFileCount bug in PageAdvanced

harryytm:
- Fix LargeFileCount able to set over 10000 at runtime in PageAdvanced

### DIFF
--- a/windirstat/Pages/PageAdvanced.h
+++ b/windirstat/Pages/PageAdvanced.h
@@ -55,7 +55,7 @@ protected:
     CStringW m_LargestFileCount;
 
     DECLARE_MESSAGE_MAP()
-    void OnEnChangeLargestFileCount();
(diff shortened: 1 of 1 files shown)


windirstat/Pages/PageAdvanced.h
```diff
@@ -53,6 +53,7 @@ class CPageAdvanced final : public CPropertyPageEx
     BOOL m_UseBackupRestore = FALSE;
     int m_ScanningThreads = 0;
     CStringW m_LargestFileCount;
+    void OnEnChangeLargestFileCount();
```
NoMoreFood: Can you move this under `DECLARE_MESSAGE_MAP()`?

harryytm: changed in latest force push

PR #281, opened 2025-09-12
Add Turkish translation (lang_tr.txt)

ileri:
This pull request adds a new Turkish translation file `lang_tr.txt`.

- Covers all strings based on the current `lang_en.txt`.
- Terminology aligned with Windows UI conventions in Turkish.
- Verified against French and English translations for context consistency.

Next step: `lang_tr.txt` can be converted into `lang_tr.bin` using the existing txt2bin tool and added to the resource script (RC file) for integration.


NoMoreFood:
Thanks! I will update the RC file.

PR #284, opened 2025-09-18
Use native header sort indicators in SortingListControl

harryytm:
Replaces manual Unicode arrow prefixing with native HDF_SORTUP and HDF_SORTDOWN flags for header sort indicators. This improves compatibility and appearance by leveraging built-in Windows header control features.
image
Before(Left) and After(Right)

Credit:Google Gemini

Prompt
-----
is there native MFC sorting indicator for the header I can use?

Response
-----
That's a fantastic question. You've hit on the most professional and elegant way to solve this problem. Yes, there is a native MFC way to handle this, and it's much simpler and more robust than manipulating strings or using custom drawing.

The Windows header control has built-in support for displaying a sorting indicator. This is controlled by a set of flags that you apply directly to a header item. By using these native flags, you completely avoid all the issues related to character pixel widths and text shifting. The operating system handles all the drawing and spacing for you, ensuring a perfect visual result on every system.

-----

### **The Native MFC Solution: `HDF_SORTUP` and `HDF_SORTDOWN`**

The key to this solution is the `HDM_SETITEM` message and the `HDITEM` structure. You can set the `fmt` member of the `HDITEM` to include the `HDF_SORTUP` or `HDF_SORTDOWN` flags. When you set these flags, Windows automatically draws a small, platform-specific arrow next to the header text.

The fix is straightforward and involves just two main steps:

1.  We will completely remove all the string manipulation code from `SortItems`. That's no longer necessary.
2.  We will modify `SortItems` to use the `HDITEM` structure to set the `HDF_SORTUP` or `HDF_SORTDOWN` flag on the currently sorted column and clear the flags from the previously sorted column.

This approach is the simplest, most correct way to handle a sorting indicator in a list control header.

#### **The C++ File to Edit**

The file you need to modify is:
`\windirstat\Controls\SortingListControl.cpp`

-----

#### **The Corrected Code**

The following code shows the new, much cleaner `SortItems` function. This single function now handles both the old and new sorting columns in a very clean way, using the native functionality of the header control.

```cpp
// \windirstat\Controls\SortingListControl.cpp

...

void CSortingListControl::SortItems()
{
    // First, sort the items in the list based on the current sorting rules.
    VERIFY(CListCtrl::SortItems([](LPARAM lParam1, LPARAM lParam2, LPARAM lParamSort) {
        const CSortingListItem* item1 = reinterpret_cast<CSortingListItem*>(lParam1);
        const CSortingListItem* item2 = reinterpret_cast<CSortingListItem*>(lParam2);
        const SSorting* sorting = reinterpret_cast<SSorting*>(lParamSort);
        return item1->CompareSort(item2, *sorting); }, reinterpret_cast<DWORD_PTR>(&m_Sorting)));

    // START MODIFICATION
    // This section updates the header's sorting indicators using native flags.
    
    CHeaderCtrl* pHeaderCtrl = GetHeaderCtrl();
    if (pHeaderCtrl == nullptr)
    {
        return;
    }

    HDITEM hditem;
    hditem.mask = HDI_FORMAT;

    // Remove the sorting indicator from the previously sorted column.
    // We must check if a column was previously sorted before trying to clear its flag.
    if (m_IndicatedColumn != -1)
    {
        pHeaderCtrl->GetItem(m_IndicatedColumn, &hditem);
        hditem.fmt &= ~(HDF_SORTUP | HDF_SORTDOWN);
        pHeaderCtrl->SetItem(m_IndicatedColumn, &hditem);
    }

    // Now, add the sorting indicator to the newly sorted column.
    pHeaderCtrl->GetItem(m_Sorting.column1, &hditem);
    hditem.fmt &= ~(HDF_SORTUP | HDF_SORTDOWN); // Clear any old flags just in case.

    if (m_Sorting.ascending1)
    {
        hditem.fmt |= HDF_SORTUP;
    }
    else
    {
        hditem.fmt |= HDF_SORTDOWN;
    }

    pHeaderCtrl->SetItem(m_Sorting.column1, &hditem);

    // Update the member variable to remember the current sorted column.
    m_IndicatedColumn = m_Sorting.column1;
    // END MODIFICATION
}
...
```

### DIFF
--- a/windirstat/Controls/SortingListControl.cpp
+++ b/windirstat/Controls/SortingListControl.cpp
@@ -152,39 +152,53 @@ void CSortingListControl::InsertListItem(const int i, CSortingListItem* item)
     VERIFY(i == CListCtrl::InsertItem(&lvitem));
 }
 
+/*
+ * Sorts the list control's items and updates the header to display the correct sorting indicator.
+ * This method reorders the list control's items based on the current sorting column and direction.
+ * It then updates the header control by using native Windows header flags (HDF_SORTUP and HDF_SORTDOWN)
+ * to display a platform-consistent sorting arrow. This approach is superior to manually
+ * changing the header text with Unicode characters, which caused visual misalignment.
+ */
 void CSortingListControl::SortItems()
 {
+    // Reorder the list items based on the current sorting criteria using a lambda comparison function.
     VERIFY(CListCtrl::SortItems([](LPARAM lParam1, LPARAM lParam2, LPARAM lParamSort) {
         const CSortingListItem* item1 = reinterpret_cast<CSortingListItem*>(lParam1);
         const CSortingListItem* item2 = reinterpret_cast<CSortingListItem*>(lParam2);
         const SSorting* sorting = reinterpret_cast<SSorting*>(lParamSort);
         return item1->CompareSort(item2, *sorting); }, reinterpret_cast<DWORD_PTR>(&m_Sorting)));
 
-    constexpr std::wstring_view sortUp = L"↑ ";
-    constexpr std::wstring_view sortDown = L"↓ ";
-
-    std::array<WCHAR, 260> text;
-    HDITEM hditem;
-    hditem.mask = HDI_TEXT;
-    hditem.cchTextMax = static_cast<int>(text.size());
+    CHeaderCtrl* pHeaderCtrl = GetHeaderCtrl();
     
-    if (m_IndicatedColumn != -1)
+    // Exit if the header control is unavailable, to prevent a null pointer crash.
+    if (pHeaderCtrl == nullptr)
     {
-        // Remove the sort indicator from the previously indicated column
-        hditem.pszText = text.data();
-        GetHeaderCtrl()->GetItem(m_IndicatedColumn, &hditem);
-        hditem.pszText = text.data() + sortUp.size();
-        GetHeaderCtrl()->SetItem(m_IndicatedColumn, &hditem);
+        return;
     }
 
-    // Copy the sort indicator to the beginning of the text buffer
-    std::wmemcpy(text.data(), m_Sorting.ascending1 ? sortUp.data() : sortDown.data(), sortUp.size());
+    HDITEM hditem;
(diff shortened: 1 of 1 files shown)


windirstat/Controls/SortingListControl.cpp
```diff
@@ -160,31 +165,41 @@ void CSortingListControl::SortItems()
...
+    pHeaderCtrl->GetItem(m_Sorting.column1, &hditem);
+    hditem.fmt &= ~(HDF_SORTUP | HDF_SORTDOWN); // Clear any old flags just in case.
+
+    if (m_Sorting.ascending1)
+    {
+        hditem.fmt |= HDF_SORTUP;
+    }
+    else
+    {
+        hditem.fmt |= HDF_SORTDOWN;
```
NoMoreFood: Good place to use the ternary operator: `hditem.fmt |= m_Sorting.ascending1 ? HDF_SORTUP : HDF_SORTDOWN`

windirstat/Controls/SortingListControl.cpp
```diff
@@ -152,6 +152,11 @@ void CSortingListControl::InsertListItem(const int i, CSortingListItem* item)
     VERIFY(i == CListCtrl::InsertItem(&lvitem));
 }
 
+/// <summary>
+/// Sorts the list control's items and updates the header to display the correct sorting indicator.
+/// <para>This method first reorders the list control's items based on the current sorting column and direction. It then updates the header control by using native Windows header flags (HDF_SORTUP and HDF_SORTDOWN) to display a platform-consistent sorting arrow.</para>
+/// <para>This approach replaces the previous method of manually prefixing the header text with Unicode characters, which caused a noticeable text shift due to inconsistent character pixel widths. The use of native flags ensures pixel-perfect alignment and full compatibility across different systems and display settings.</para>
+/// </summary>
```
NoMoreFood: We do not use this commenting style anywhere else in the program so just stick to // or /* ... */

NoMoreFood:
I really like this change.  See my comments, but otherwise looks great.

PR #288, opened 2025-10-01
Fix auto-size truncation and adjust column padding

harryytm:
- Increase GetSubItemWidth measurement width to 3500 to prevent truncation of very long file name
- Change column padding from 5px to 3px
- Add 3px padding to Auto Resize Columns function to prevent truncation

### DIFF
--- a/windirstat/Controls/OwnerDrawnListControl.cpp
+++ b/windirstat/Controls/OwnerDrawnListControl.cpp
@@ -508,7 +508,7 @@ bool COwnerDrawnListControl::IsShowSelectionAlways() const
 int COwnerDrawnListControl::GetSubItemWidth(COwnerDrawnListItem* item, const int subitem)
 {
     CClientDC dc(this);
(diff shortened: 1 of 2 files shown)


windirstat/Controls/OwnerDrawnListControl.cpp
```diff
@@ -508,7 +508,7 @@ bool COwnerDrawnListControl::IsShowSelectionAlways() const
 int COwnerDrawnListControl::GetSubItemWidth(COwnerDrawnListItem* item, const int subitem)
 {
     CClientDC dc(this);
-    CRect rc(0, 0, 1000, 1000);
+    CRect rc(0, 0, 3500, 20);
```
harryytm, line 511: Tested 3400px required for 1 nested level (under a root item and an item) with 254 characters long file name on the tree, add 100px as safe margin for extreme case of heavily nested file path and very long file name.

PR #289, opened 2025-10-02
Fix columns autosize (divider double click) to respect header text

harryytm:


windirstat/Controls/OwnerDrawnListControl.cpp
```diff
@@ -530,6 +531,65 @@ int COwnerDrawnListControl::GetSubItemWidth(COwnerDrawnListItem* item, const int
...
+    szHeaderText[_countof(szHeaderText) - 1] = _T('\0');
+
+    const int padding = 9; // Padding adjustment.
+    int totalWidth;
+
+    HTHEME hTheme = ::OpenThemeData(pHeaderCtrl->GetSafeHwnd(), _T("HEADER"));
+
+    if (hTheme)
+    {
+        CRect rectThemeText;
```
NoMoreFood: Was this generated by AI or did you find this was actually necessary?

windirstat/Controls/OwnerDrawnListControl.cpp
```diff
@@ -530,6 +531,65 @@ int COwnerDrawnListControl::GetSubItemWidth(COwnerDrawnListItem* item, const int
...
+
+    if (!pHeaderCtrl)
+    {
+        return 0;
+    }
+
+    CClientDC dc(pHeaderCtrl);
+    CFont* pOldFont = dc.SelectObject(pHeaderCtrl->GetFont());
+
+    TCHAR szHeaderText[256];
```
NoMoreFood: We only use WCHARs in WinDirStat.

windirstat/Controls/OwnerDrawnListControl.cpp
```diff
@@ -530,6 +531,65 @@ int COwnerDrawnListControl::GetSubItemWidth(COwnerDrawnListItem* item, const int
...
+    hdItem.cchTextMax = _countof(szHeaderText);
+
+    if (!pHeaderCtrl->GetItem(column, &hdItem))
+    {
+        dc.SelectObject(pOldFont);
+        return 0;
+    }
+
+    // Explicitly null-terminate the string.
+    szHeaderText[_countof(szHeaderText) - 1] = _T('\0');
```
NoMoreFood, line 561: You already zero this out earlier.  It would seem more logical to set cchTextMax to _countof(szHeaderText) - 1 if we're looking to avoid overflow.

windirstat/Controls/OwnerDrawnListControl.cpp
harryytm: I can change that

windirstat/Controls/OwnerDrawnListControl.cpp
harryytm: If the else block is to keep, this line is needed or the IntelliSense in VS2022 would mark that as warning.
```
Severity Code   Description                                             Project         File                                                                    Line
Warning  C6054  String 'szHeaderText' might not be zero-terminated.     windirstat      D:\GitHub\windirstat\windirstat\Controls\OwnerDrawnListControl.cpp      586 
````

windirstat/Controls/OwnerDrawnListControl.cpp
harryytm: Which block do you prefer? The True block or the False block?

I keep questioning Gemini is the fallback block really necessary, but it keep convince me to keep it to handle edge cases.

I am finding a way to trigger hTheme = false in Windows 10 to test if the fallback block is needed, but if it don't hurt the accuracy of the header width it returns. I am also both version (True side and False side) to see if any difference.

If there is no difference on affecting the accuracy I would happy to make the whole function as simple as possible.

All I want was autosize (divider double click) to respect header text width without truncation.

NoMoreFood: The false block.

harryytm:
I have tested the build.

NoMoreFood:
Closing in favor of https://github.com/windirstat/windirstat/commit/2f2d5fc718128f8908f0139f886d031f0b552d51

Let me know if that doesn't do the trick and I'll reopen this one.

harryytm:
> Closing in favor of [2f2d5fc](https://github.com/windirstat/windirstat/commit/2f2d5fc718128f8908f0139f886d031f0b552d51)
> 
> Let me know if that doesn't do the trick and I'll reopen this one.

2f2d5fc718128f8908f0139f886d031f0b552d51 is almost perfect perfect since it much cleaner and elegant, but the problem is the last column which would fill the rest of available space, when couple with "Automatically resize columns" enabled, that would cause the horizontal scroll bar to show in the treelist view and the user have to scroll all the way to the right to manually resize or reach the divider to shrink the column size, which is not a very good user experience.

Furthermore in order to disable that fill function of LVSCW_AUTOSIZE_USEHEADER, might need many lines of code just to disable the last column fill behavior, which as I can see the implementation largely the same as COwnerDrawnListControl::GetHeaderWidth(const int column) in mp PR.

I really love to accept the LVSCW_AUTOSIZE_USEHEADER approach but it have its side effect which I can't accept, and it is up to you decide using 2f2d5fc718128f8908f0139f886d031f0b552d51 or using GetHeaderWidth() I came up with in this closed PR.

PR #290, opened 2025-10-03
Auto Reduce Frequent Reallocations In Path Construction

harryytm:
- Implement self-tuning cache for maximum path depth to auto reduce memory reallocation overhead in path construction
- Change pathParts vector initialization from resize(8) to reserve(MaxPathPartsSize)
- Use std::atomic<size_t> with Compare-and-Swap (CAS) logic for MaxPathPartsSize to ensure thread safety during concurrent scanning
- Reset MaxPathPartsSize cache to default at the start of every new scan

### DIFF
--- a/windirstat/DirStatDoc.cpp
+++ b/windirstat/DirStatDoc.cpp
@@ -1517,6 +1517,9 @@ void CDirStatDoc::StartScanningEngine(std::vector<CItem*> items)
     // Stop any previous executions
     StopScanningEngine();
 
+    //Reset Path Construction Cache Size to Default
+    CDirStatApp::Get()->ResetMaxPathPartsSizeCache();
+
     // Address currently zoomed / selected item conflicts
     const auto zoomItem = GetZoomItem();
     for (const auto& item : std::vector(items))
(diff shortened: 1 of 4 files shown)


NoMoreFood:
This PR will end up reducing the compiler optimization potential for pathParts.resize(8) which is a reasonable high-average depth size.  I have no intent of allow the users to dynamically update this so the additional maintenance burden of the code to make it more accessible/settable is not worth it.

harryytm:
The main reason I made this PR was pathParts.resize(8) cause WinDirStat to crash when starting a scan, it would be a simple fix just to replace it with pathParts.reserver(8), I digged deeper to find out the peak size of pathParts was 23 during my test, and majority of the time was around 8, by using VTRACE to track the peak and current of the size of pathParts, and I give it a go to have fun with playing with the code try to make it automatically raise the peak size.

I really happy for your time reviewing the code I made and thanks for your feedback.

NoMoreFood:
Oh geez.  You are 100% correct. It should have been reserve(8), not resize(8).  I will fix.

PR #302, opened 2025-10-17
Enhance and refactor external localization loading

harryytm:
- Add forward-compatible support for modern BCP 47 language code (e.g., lang_en-US.txt) filename for external localization files, would try original ISO 639-1 language code first (e.g., lang_en.txt), then BCP 47 language code
- Refactors the external localization file checking and loading logic into a separate function `LoadExternalLang`

### DIFF
--- a/windirstat/Localization.cpp
+++ b/windirstat/Localization.cpp
@@ -94,13 +94,8 @@ std::vector<LANGID> Localization::GetLanguageList()
 
 bool Localization::LoadResource(const WORD language)
 {
-    FinderBasic finder;
-    const std::wstring lang = GetLocaleString(LOCALE_SISO639LANGNAME, language);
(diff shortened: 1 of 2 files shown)


harryytm:
Reason for adding BCP 47 language code is ISO 639-1 language code don't have a code for Chinese Traditional (Hong Kong).
Tested function correctly.

harryytm:
Before this change, setting the language to Chinese (Traditional, Hong Kong) or Chinese (Simplified) WinDirStat would try to find and load lang_zh.txt during startup.

After this change, WinDirStat would try to find and load:
- lang_zh.txt then lang_zh-CN.txt for Chinese (Simplified)
- lang_zh.txt then lang_zh-HK.txt for Chinese (Traditional, Hong Kong)
- loading of internal language resources will not be affected if no external language file found

PR #306, opened 2025-10-24
Fix incorrect French translation for "Empty folder"

thomaslevesque:
The old translation treated "empty" as an adjective, when it's actually a verb in that context.


NoMoreFood:
Thanks!

PR #309, opened 2025-10-28
Add github workflow to easily build Windirstat between releases

newrecycle:
Simple workflow that builds windirstat. Making for easier testing of commits between releases without the need to pull and build locally.


NoMoreFood:
Since we have automated integration with releasing to Winget (even for betas), I don't want anything like this automatically running.  I appreciate the contribution though.

Issue #313, opened 2025-11-04
Darkmode Feedback


harryytm:
First of all @NoMoreFood you have done amazing work adding long asked Dark Mode to WInDirStat, it looks amazing to me and I am sure I would use it all the time. Thanks for implementing Dark Mode and keep up the great work.

The following text labels in Options dialog are in black in Dark Mode, still need to refine the colors to make it readable.

<img width="594" height="452" alt="Image" src="https://github.com/user-attachments/assets/efbbf76d-4f32-4a4d-b1ce-5cb8795190dc" />
<img width="594" height="452" alt="Image" src="https://github.com/user-attachments/assets/1d0c5ec3-6eb1-4069-bc29-96a0be95afbb" />
<img width="594" height="452" alt="Image" src="https://github.com/user-attachments/assets/c99e9c32-312b-4d3f-9090-88b1d9abbf68" />
<img width="594" height="452" alt="Image" src="https://github.com/user-attachments/assets/b187a3c2-4a24-4384-a1d7-bf6891f650e9" />
<img width="594" height="452" alt="Image" src="https://github.com/user-attachments/assets/d240f61a-b487-4536-ae4b-aa5a972f0bc9" />
<img width="594" height="452" alt="Image" src="https://github.com/user-attachments/assets/c1afa77f-58a8-43ac-b155-ad6fb8254723" />

NoMoreFood:
Thanks.  I'll look into it.

harryytm:
Created PR #314 for fixing the issues by guidance of Google Gemini, tested fix the problem.

harryytm:
Text color of the NTFS compressed files or folders need to use a lighter blue to have better contrast with the dark background.
Windows 10 by default uses RGB 58,153,232 (measured with dropper tool in MS paint) as text color for NTFS compressed files or folders in dark mode.
Given that if you prefer to align with Windows dark mode color scheme.

<img width="1280" height="49" alt="Image" src="https://github.com/user-attachments/assets/b7f7f65e-7857-4e14-a22a-015f3fddd923" />

Windows 10
<img width="632" height="49" alt="Image" src="https://github.com/user-attachments/assets/73c0edd1-4dec-4eb1-bbd7-9031a189acff" />

NoMoreFood:
I changed it to use the Windows 10 default (for both light and dark).  Can you see if that's a good compromise for both?  If not, we can adjust it whether in dark mode or not.

harryytm:
[...]

Issue #106, opened 2024-10-22
Dark Mode


aboe026:
It would be great if WinDirStat had a dark mode/theme to make it easier on the eyes when using it late at night.

NoMoreFood:
I would be supportive of adding this feature but would really love someone to play around and see how far they can get.

hueldoeu:
yes. dark mode is a must have. i have been banned on airbnb for invalid reasons but they still don't offer a dark mode therefore i must refuse on airbnb. dark mode is a useful setting.

whoever helps me with using their Govt ID to get a verified and non-banned airbnb account and keep all the money i earn. i just like to be a hospitable person.

Neustradamus:
Very important feature!

Denyerec:
I remember once upon a time long ago when you could change the colours in Windows, and all apps would honour them... Absolutely mad that developers have to create a customised scheme for something that should be governed by the OS :(

Bleauhaus:
@NoMoreFood no skills to help here but I offer a strong vote for this

clngr:
Great feature to have!

b1k3rdude:
Just adding my 2 cents, I would also like the option of having Darkmod please :-)

squeak9:
It's officially in the works now!! https://github.com/windirstat/windirstat/commit/53ff23d075393a812cf1ac8283592d774363daaa

Screenshots: https://github.com/windirstat/windirstat/issues/313

I guess you have to build it yourself right now but sounds like a new beta is coming soon: https://github.com/windirstat/windirstat/issues/313#issuecomment-3489002791

NoMoreFood:
Can you please validate this build addresses your request / issue?

https://github.com/windirstat/windirstat/releases/tag/beta/v2.4.0/2025-11-07

Neustradamus:
Good job, thanks to @harryytm, @NoMoreFood.

Linked to:
- https://github.com/windirstat/windirstat/issues/106
- https://github.com/windirstat/windirstat/issues/313
- https://github.com/windirstat/windirstat/pull/314
- https://github.com/windirstat/windirstat/pull/317
- https://github.com/search?q=repo%3Awindirstat%2Fwindirstat+Dark+Mode&type=commits

Bleauhaus:
Excellent !!
My Dark Mode Campaign conquers yet another system! Good Job @Decision
Makers for putting in the effort to do this. I for one have worsening
Eyesight issues and this makes a big difference for me, And It Looks cool
[...]

PR #314, opened 2025-11-04
Fix checkbox label color in dark mode

harryytm:
Credit: Google Gemini guided patch

## Prompt

The labels of the radio buttons already in correct color before changing anything

## Response

That's very helpful information\! If the **radio button labels** are already in the correct light color, it means your existing code is successfully handling:

1.  **Radio Button Styles:** `BS_RADIOBUTTON` and `BS_AUTORADIOBUTTON`.
2.  **Group Box Styles:** `BS_GROUPBOX` (which often contains the radio buttons).

This confirms that the existing logic for handling these specific button types in `DarkMode::AdjustControls` is correct for setting up the text color.

Since the **Checkbox labels are black**, but the **Radio Button labels are correct**, this points to a specific gap in how you are applying the theme override for the **Checkbox control styles**.

-----

## 🧐 Focused Analysis: Checkbox vs. Radio Button Logic

Let's look closely at your existing `DarkMode::AdjustControls` logic to identify where the radio button styles are handled, and how we need to extend that to include the checkbox styles.

### 1\. The Current Successful Logic (Radio Buttons)

Your code that successfully allows radio buttons to use the correct color is this:

```cpp
// From DarkMode::AdjustControls
if (className == WC_BUTTON)
{
    const auto style = GetWindowLong(hWnd, GWL_STYLE) & BS_TYPEMASK;
    if (style == BS_GROUPBOX || style == BS_RADIOBUTTON || style == BS_AUTORADIOBUTTON)
    {
        // This line is likely disabling the Windows Theme for these controls,
        // allowing the WM_CTLCOLORSTATIC/WM_CTLCOLORBTN message to apply the color.
        SetWindowTheme(hWnd, _darkModeEnabled ? L"DarkMode_Explorer" : L"Explorer", L""); 
    }
// ...
```

The key to the radio button success is that the control's style (`BS_RADIOBUTTON` or `BS_AUTORADIOBUTTON`) is caught in this `if` block, and the subsequent `SetWindowTheme` call (with the empty string parameter `L""`) effectively **removes the visual theme**, forcing the GDI drawing mode which respects the color set by your `OnCtlColor` function.

### 2\. The Missing Logic (Checkboxes)

The issue is that if the control style is **`BS_CHECKBOX`** or **`BS_AUTOCHECKBOX`**, it currently falls through this block and continues to use the default themed drawing, which forces the black text color.

To fix the black checkbox labels, you must include the checkbox styles in that same block of code to remove the theme.

-----

## 💡 The Minimal Fix: Modifying `DarkMode::AdjustControls`

You need to update the `if` condition in `DarkMode::AdjustControls` to explicitly include the two common checkbox styles:

#### Required Code Change

Locate this line in your `DarkMode::AdjustControls` function:

```cpp
// Original condition
if (style == BS_GROUPBOX || style == BS_RADIOBUTTON || style == BS_AUTORADIOBUTTON)
```

And update it to this, ensuring the checkbox styles are also included:

```cpp
// Corrected condition
if (style == BS_GROUPBOX || style == BS_RADIOBUTTON || style == BS_AUTORADIOBUTTON || 
    style == BS_CHECKBOX || style == BS_AUTOCHECKBOX) // <--- ADDED CHECKBOX STYLES
{
    // ... the SetWindowTheme(hWnd, ..., L"") call follows, which forces the color change
    SetWindowTheme(hWnd, _darkModeEnabled ? L"DarkMode_Explorer" : L"Explorer", L"");
}
```

By adding `BS_CHECKBOX` and `BS_AUTOCHECKBOX`, you ensure that checkboxes are also theme-disabled, which will force their labels to be drawn using GDI, allowing your existing `CTLCOLOR_STATIC` (or similar) logic in `OnCtlColor` to correctly paint the text label white.

-----

Would you like to confirm the exact lines in the `DarkMode::AdjustControls` function so you can apply this change?

### DIFF
--- a/windirstat/DarkMode.cpp
+++ b/windirstat/DarkMode.cpp
@@ -146,7 +146,8 @@ void DarkMode::AdjustControls(const HWND hWnd)
         if (className == WC_BUTTON)
         {
             const auto style = GetWindowLong(hWnd, GWL_STYLE) & BS_TYPEMASK;
-            if (style == BS_GROUPBOX || style == BS_RADIOBUTTON || style == BS_AUTORADIOBUTTON)
+            if (style == BS_GROUPBOX || style == BS_RADIOBUTTON || style == BS_AUTORADIOBUTTON ||
+                style == BS_CHECKBOX || style == BS_AUTOCHECKBOX)
             {
                 SetWindowTheme(hWnd, _darkModeEnabled ? L"DarkMode_Explorer" : L"Explorer", L"");
             }


NoMoreFood:
I had already submitted something to only affect the "normal" buttons since it didn't look like the theme was affecting anything other than that.  See if what I already committed makes it better for you please.

harryytm:
Your solution is better and tested resolved the issues, closing this PR.

Neustradamus:
Good job, thanks to @harryytm, @NoMoreFood.

Linked to:
- https://github.com/windirstat/windirstat/issues/106
- https://github.com/windirstat/windirstat/issues/313
- https://github.com/windirstat/windirstat/pull/314
- https://github.com/windirstat/windirstat/pull/317
- https://github.com/search?q=repo%3Awindirstat%2Fwindirstat+Dark+Mode&type=commits

Issue #340, opened 2025-12-17
Hardlink Support Feedback


harryytm:
@NoMoreFood I welcome the implementation of hardlink support since acbb6cdb62c5127e3d6e6681572c476f1376480c, but I hate to say something you might not like, that the implementation have a major logical flaw which render the scan result useless. I have waited for almost a week after that commit, hope you might notice and fix this critical design flaw, but still don't see it coming, other than commits for performance tweaks and refactoring in the past 10 days, until I see you recently inclement the version number, hence I am posting this issue to raise my concern.

The core function of WinDirStat is to show the size of files on Three views, but with the initial hardlink support, the physical size no longer accurate, due to the fact that physical size of all the files that referenced via hardlink are now marked as zero in size and merged into a single virtual item on the tree view `<Hardlinks>`. Using logical size as alternative is not a viable either, since it is not the actual disk space occupied by the file on the disk.

Take my C drive as example, before the hardlinks support, the total used physical size on the tree view is 231 GiB, and Windows report the drive used space is 228 GiB, which means the actual `<Unknown>` should be -3 GiB, but after acbb6cdb62c5127e3d6e6681572c476f1376480c, on the same disk at the same time the `<Unknown>` in the scan result reported 7.8 GiB, with 8.8 GiB of `<Hardlinks>`, which is a size that don't make any sense at all.

I urge you please rethink the implementation of Hardlink Support, otherwise WinDirStat would become a unusable tool for finding space hog files, which I don't whish to see. Please reconsidered the suggested solution I have posted in #108 on August as follows:

> @NoMoreFood I have an idea to represent the shared disk space of the hardlink, which is have a new optional entry called "Hardlink Shared Space", the idea of this entry is for showing the disk space saved by hardlink, for example 1GiB of data referenced by 5 hardlinks would be 1GiB x 4 = 4 GiB as "Hardlink Shared Space", and this number should be add back to "Unknown" to balance the negative number cause by hardlink, which currently shown as 0 when that happen.
> 
> This idea is borrowed from the report of Dism /Online /Cleanup-Image /AnalyzeComponentStore, which shows the used disk of space Component Store shared with Windows.
> 
> ```
> Deployment Image Servicing and Management tool
> Version: 10.0.19041.3636
> 
> Image Version: 10.0.19045.6216
> 
> [==========================100.0%==========================]
> 
> Component Store (WinSxS) information:
> 
> Windows Explorer Reported Size of Component Store : 12.80 GB
> 
[...]

PR #344, opened 2025-12-26
Hardlinks navigation control refinements

harryytm:
- change backlink symbol to ⫘

### DIFF
--- a/windirstat/Item.cpp
+++ b/windirstat/Item.cpp
@@ -148,7 +148,7 @@ std::wstring CItem::GetText(const int subitem) const
         }
         if (IsTypeOrFlag(IT_HLINKS_FILE))
         {
-            return std::wstring(L"🔗 ") + FormatBytes(GetSizePhysical());
+            return std::wstring(L"⫘ ") + FormatBytes(GetSizePhysical());
         }
         return FormatBytes(GetSizePhysical());
 


harryytm:
@NoMoreFood Thanks for implementating the backlink function in 8df1a8c831288142d57c2ef62651e59596271c17 which drastricly improve usability of hardlink viewing function, just need a little bit of UI controls change to make the hardlinks navigation experience perfect and more intuitive for user to read the hardlink paths in one go, instead of rely on status bar to see where the path of the hardlink is on each hardlink item.

NoMoreFood:
I like the idea and was originally considering this, but still don't like the fact it widens the column so much since the path is not tiered.  I still think I like it though so I think I'm good with it.

I like differentiating the hard link icon as well, but why wouldn't we just replace the normal icon with something like this one instead of having it have two icons?

NoMoreFood:
Yeah, what I think I'd like to do is make the file icon itself the actual icon and then keep the physical column size clickable (not the name).   And we can switch the physical column icon to be the one you found. I can see folks wanting to click on the file name (to select it) and do things like copy and paste the path or delete it without having to bounce back and forth.

harryytm:
> I like the idea and was originally considering this, but still don't like the fact it widens the column so much since the path is not tiered. I still think I like it though so I think I'm good with it.

good to hear you liked it

> I like differentiating the hard link icon as well, but why wouldn't we just replace the normal icon with something like this one instead of having it have two icons?

I agree with 2 icons on the name looks messy, but the nature of this PR is like feature request prototype instead of being final, so I try to minimize the change, plus my C++ reading is still not very good yet, so if you know how to make this change please do it

> Yeah, what I think I'd like to do is make the file icon itself the actual icon and then keep the physical column size clickable (not the name). And we can switch the physical column icon to be the one you found. I can see folks wanting to click on the file name (to select it) and do things like copy and paste the path or delete it without having to bounce back and forth.

agreed with that, change this PR to what you like if you want, I still have 2 more working day before my next off

I have another feature wish as well, see if it is possible to add the hardlinks references as I mentioned in #340 aka the `⧉ 196 KiB (2)` thing as well, thanks.

windirstat/Item.cpp
```diff
@@ -148,7 +148,7 @@ std::wstring CItem::GetText(const int subitem) const
         }
         if (IsTypeOrFlag(IT_HLINKS_FILE))
         {
-            return std::wstring(L"🔗 ") + FormatBytes(GetSizePhysical());
+            return std::wstring(L"⫘ ") + FormatBytes(GetSizePhysical());
```
NoMoreFood, line 151: You think this one is better?  I'm sort of indifferent, but I stopped using 🔗 for the reparse points so I thought it might look good here.

harryytm: I love to use 🔗 but it not quite looks good on the UI and doesn't looks like a link symbol at first glance for some reason, but rather look like a bandage or pill instead, due to the chain links holes being filled in both white and black mode
image
however ⫘ looks cleaner and immediately understandable as a link by users
image

PR #355, opened 2026-01-01
Refactor CMessageBox to streamline calls

harryytm:
- Introduced three Show() static overloads to streamline CMessageBoxDlg calls and mimic the calling style of MessageBox():
    - Standard: with additional optional parameter to control initial window size.
    - With Checkbox: with additional checkbox support for persistent options like "Don't show again."
    - File Delete Dialog: additionally accepts vector-based list data for displaying multiple paths
- Added struct WdsMessageBoxResult to allow returning both button IDs and checkbox states
- Optimized parameters arrangement and defaults to help shorten calls by internalizing common values like application titles and parent handles
- Refactored call sites using the IIFE pattern to improve code readability

### DIFF
--- a/windirstat/Controls/FileDupeControl.cpp
+++ b/windirstat/Controls/FileDupeControl.cpp
@@ -43,11 +43,18 @@ void CFileDupeControl::ProcessDuplicate(CItem* item, BlockingQueue<CItem*>* queu
         m_showCloudWarningOnThisScan = false;
 
         // Show warning dialog
-        CMessageBoxDlg dlg(Localization::Lookup(IDS_DUPLICATES_WARNING), Localization::LookupNeutral(AFX_IDS_APP_TITLE),
-            MB_OK | MB_ICONINFORMATION, this, {}, Localization::Lookup(IDS_DONT_SHOW_AGAIN), false);
-        if (dlg.DoModal() == IDOK && dlg.IsCheckboxChecked())
+        if (COptions::SkipDupeDetectionCloudLinks && item->IsTypeOrFlag(ITRP_CLOUD) && m_showCloudWarningOnThisScan)
         {
-            COptions::ShowDupeDetectionCloudLinksWarning = false;
+            m_showCloudWarningOnThisScan = false;
+
+            // Show warning and immediately "unpack" the returned structure
+            if (const auto [nID, isChecked] = CMessageBoxDlg::Show(Localization::Lookup(IDS_DUPLICATES_WARNING),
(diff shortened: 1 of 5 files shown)


harryytm:
Brainstorming how to further simplify CMessageBox calls to improve code readability, hence change this PR to draft.

NoMoreFood:
@harryytm I know this (and those other PRs) have been out there and never been marked as ready.  Are these still applicable?

harryytm:
Still need more time to think about how to arrange the parameters after you added the autosize function.
WOuld be lovely to have some input from you on how you think about the intentions of this PR, my aim was to reduce or remove the boilerplate calls from the code and make it more readable and maintainable.

windirstat/Controls/FileDupeControl.cpp
```diff
@@ -43,12 +43,17 @@ void CFileDupeControl::ProcessDuplicate(CItem* item, BlockingQueue<CItem*>* queu
         m_showCloudWarningOnThisScan = false;
 
         // Show warning dialog
-        CMessageBoxDlg dlg(Localization::Lookup(IDS_DUPLICATES_WARNING), Localization::LookupNeutral(AFX_IDS_APP_TITLE),
-            MB_OK | MB_ICONINFORMATION, this, {}, Localization::Lookup(IDS_DONT_SHOW_AGAIN), false);
-        if (dlg.DoModal() == IDOK && dlg.IsCheckboxChecked())
-        {
-            COptions::ShowDupeDetectionCloudLinksWarning = false;
-        }
+        [&]() {
```
NoMoreFood: What is the purpose of the lambda in this case?

harryytm: For isolating `result`, will be replaced with a block scope.

harryytm: Asked Gemini Fast for a better solution to eliminate temporary variable, looks like  this meets my intension even better.

```cpp
if (COptions::SkipDupeDetectionCloudLinks && item->IsTypeOrFlag(ITRP_CLOUD) && m_showCloudWarningOnThisScan)
{
    m_showCloudWarningOnThisScan = false;

    // Show warning and immediately "unpack" the returned structure
    if (const auto [nID, isChecked] = CMessageBoxDlg::Show(
            Localization::Lookup(IDS_DUPLICATES_WARNING), 
            Localization::Lookup(IDS_DONT_SHOW_AGAIN), 
            false, MB_OK | MB_ICONINFORMATION, this); 
        nID == IDOK && isChecked)
    {
        COptions::ShowDupeDetectionCloudLinksWarning = false;
    }

    return;
}
```

What do you think?

PR #373, opened 2026-01-07
Fix extension search regex

harryytm:


NoMoreFood:
Your fix keyed me into the true problem -- I was assuming SearchWholePhrase was alse instead of saving it off and overriding it like the other variable.  Instead of trying to save everything off, I converted ProcessSearch() to take them as parameters so we do not worry about persistence.

PR #384, opened 2026-01-16
Update lang-ko

VenusGirl:
Update Korean


VenusGirl (commented):
Korean language correction update

PR #399, opened 2026-01-26
Add compile time temporary language file to gitignore

harryytm:


.gitignore
```diff
@@ -8,3 +8,4 @@
 /enc_temp_folder/
 /.git/
 .vs/
+windirstat/res/lang_combined.txt
```
NoMoreFood, line 11: The PowerShell script should be removing this.  Is it not getting removed for you:

`If ($File.Name -eq 'lang_combined.txt') { Remove-Item $TxtFile -Force }`

harryytm: I understand the compile script would remove this after the compile, but just being annoying this would momentary appear in GitHub Desktop changed files during compile, and have the risk to contaminate the commit by accident, I highly suggest adding it to .gitignore, or alternatively use .bin extension for this temporary file.

PR #411, opened 2026-02-08
Add silent mode for command-line execution

arunvjs:
This change introduces a silent mode for WinDirStat, allowing it to be run from the command line to scan a directory and save the results to a CSV file without displaying a graphical user interface.

**Purpose:**
- To enable automation and scripting of WinDirStat scans.
- To allow for generating directory analysis reports as part of automated processes or batch jobs.

**Usage:**
The new ``/silent`` command-line argument is used to trigger this mode. When running in silent mode, a path to scan and an output CSV file must be provided.

**Syntax:**
windirstat.exe /silent <directory_to_scan> <path_to_output.csv>

**Example:**
windirstat.exe /silent "C:\Program Files" "C:\reports\windirstat_report.csv"

This enhancement allows users to leverage WinDirStat's capabilities in a more flexible way, particularly for system administrators and power users who need to generate reports without manual intervention.

and changed the version number from 2.5.3 to 2.5.4 in Version.h

### DIFF
--- a/windirstat/Version.h
+++ b/windirstat/Version.h
@@ -35,7 +35,7 @@
 
 #define PRD_MAJVER                  2 // major product version
 #define PRD_MINVER                  5 // minor product version
-#define PRD_PATCH                   3 // patch number for product
+#define PRD_PATCH                   4 // patch number for product
 #define PRD_BUILD                   GIT_COUNT // build number for product
 #define FILE_MAJVER                 PRD_MAJVER // major file version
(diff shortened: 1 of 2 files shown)


NoMoreFood:
How is this different from running `WinDirStat_x64.exe /SaveToCSV v:\OutFile.csv C:\PathToScan` which already runs silently.

NoMoreFood:
@arunvjs Thoughts on my question?

NoMoreFood:
Closing; will reopen depending if a response is received.

PR #417, opened 2026-02-13
Add hardcode UI strings to language files

harryytm:


NoMoreFood:
Are you seeing "This PC" actually show up?  This was just a fallback value in case the OS lookup fails?

harryytm:
I knew it is a fallback, it is just me don't like this single language (English) hardcode UI string hanging around in the code.

PR #421, opened 2026-02-19
Address tools being disabled regression

harryytm:
Address the following tools being disabled regression
- Defragment Volume
- Check Disk

### DIFF
--- a/windirstat/MainFrame.cpp
+++ b/windirstat/MainFrame.cpp
@@ -1144,8 +1144,8 @@ void CMainFrame::UpdateToolsMenu(CMenu* menu)
     // Get available local drives and conditionally enable based on elevation
     const auto drives = GetDriveList({DRIVE_FIXED, DRIVE_REMOVABLE, DRIVE_RAMDISK});
     SetMenuItem(menu, shadowCopyPos, IsElevationActive() && !drives.empty());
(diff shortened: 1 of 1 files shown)


NoMoreFood:
This was intentional.  These commands support elevating upon execution since they are external processes. What are you seeing?

harryytm:
What I saw was those two tools always being disabled when WDS is running in admin, the reason is if IsElevationActive() is true, IsElevationAvailable() would be false.

```cpp
bool IsElevationAvailable() noexcept
{
    static const auto result = []() noexcept -> bool
    {
        if (IsElevationActive()) return false;
        
        SmartPointer<HANDLE> token(CloseHandle);
        TOKEN_ELEVATION_TYPE elevationType;
        DWORD size = sizeof(TOKEN_ELEVATION_TYPE);
        if (OpenProcessToken(GetCurrentProcess(), TOKEN_QUERY, &token) == 0 ||
            GetTokenInformation(token, TokenElevationType, &elevationType, sizeof(elevationType), &size) == 0)
        {
            return false;
        }
        
        return elevationType == TokenElevationTypeLimited;
    }();
    
    return result;
}
```

NoMoreFood:
Okay, in that case, can you switch it to IsElevationAvailable() || IsElevationActive()

harryytm:
> Okay, in that case, can you switch it to IsElevationAvailable() || IsElevationActive()

Amended and tested working.

PR #424, opened 2026-02-19
Fix Chinese Simplified

harryytm:
- fix IDS_RAMUSAGEs missing {} in the string and correcting translation

Before
image

After
image

Bug reference: 8ad84510474105b564bc9dcf8dcdfc942dc556c2


harryytm:
# Extended explanations generated by Google Gemini - Fast
The shift from `内存使用情况` (Memory Usage Situation) to `内存占用` (Memory Occupancy) in Simplified Chinese software localization is not merely a stylistic choice; it represents a fundamental alignment with the technical vocabulary and cognitive expectations of users in Mainland China. To understand why `内存占用` is considered "native" while `内存使用情况` feels like a literal, non-standard translation, one must examine the influence of the Windows ecosystem and the linguistic nuances of technical terminology within the PRC.

### The Standardizing Power of Windows Task Manager

The primary reason `内存占用` feels native to Mainland Chinese users is the standardizing influence of Microsoft Windows. Since the mid-1990s, the Windows operating system has served as the primary interface through which Mainland users interact with computer architecture. In the Simplified Chinese version of Windows, specifically within the **Task Manager (任务管理器)**, the term used to describe a program's consumption of physical RAM is almost exclusively **"占用"**.

When a user opens their system monitor to diagnose a slow computer, they look for the "内存" (Memory) column to see which process is "占用" (occupying) the most resources. Consequently, this specific two-character term has become a "mental landmark." For a third-party utility like WinDirStat to feel integrated into the system, it must speak the same technical language as the host OS. Using the OS-standard term validates the software’s authority and makes it feel like a professional system tool rather than an external, translated application.

### Linguistic Efficiency and Technical Tone

In the context of software UI design in the PRC, brevity is often equated with professionalism.

* **内存使用情况 (Nèicún Shǐyòng Qíngkuàng):** This phrase translates literally to "Memory Use Situation" or "Circumstances of Memory Usage." While grammatically correct, it is wordy and descriptive. In Chinese linguistic culture, adding "情况" (situation/circumstances) at the end of a noun phrase often makes it sound like a title in a general report or a sentence in a textbook. It lacks the "punch" required for a functional UI label.
* **内存占用 (Nèicún Zhànyòng):** This term is concise and action-oriented. It combines the noun for memory with a specific verb-noun that means "to take up space" or "occupy." It sounds like an engineering status report.

For Mainland users, the word **"用量"** (Amount used) or **"使用情况"** is frequently associated with consumables that are permanently spent, such as water, electricity, or mobile data. However, computer memory is a finite space that is occupied and then vacated. The term **"占用"** perfectly captures this spatial logic of RAM allocation, making it the more "logical" and native choice for a tool that analyzes disk and memory space.

### The "Translation Ghost" in `内存使用情况`

The string `内存使用情况` is often identified by native users as a "Translation Ghost"—a term that exists only because it was translated literally from the English "Memory Usage."

When a translator looks at the English source `Memory Usage`, the most direct dictionary translation for `Usage` is `使用` (use) or `使用情况` (usage circumstances). However, direct translation often fails to account for how a local culture has naturally named a concept in its own technical evolution. Because Mainland China developed a robust, localized tech ecosystem very early on, it established its own "tech-slang" and formal standards.

Using `内存使用情况` signals to a Mainland user that the software was likely translated by someone using a dictionary or a machine translation tool without knowledge of the local Windows environment. It creates a "foreign" feel, similar to how an English speaker might feel if a program used the term "Computer Brain Storage" instead of "Hard Drive." It is technically descriptive, but it isn't what people actually say.

### Functional Consistency in System Utilities

System utilities like WinDirStat are used by "power users"—individuals who are often more sensitive to terminology than the average user. These users expect a high degree of precision. In the PRC's technical community, `占用` is the specific term used in coding documentation, system administration, and performance benchmarking.

By adopting `内存占用`, the software moves away from being a "description of a situation" to a "data point of occupancy." This subtle shift in tone aligns with the user's objective: they aren't looking to understand the "situation" of their memory; they are looking to see what is "occupying" it so they can free it up.

In conclusion, `内存占用` is native because it is the vocabulary of the environment the user lives in. It respects the standard set by the operating system, it utilizes the spatial logic inherent in the Chinese understanding of resource allocation, and it avoids the clunky, descriptive nature of literal translation. For a project like WinDirStat, this level of localization is what transforms a "translated" app into a "native" experience.

PR #428, opened 2026-02-22
Change drive selection UI to visually select relevant drives when All is chosen

### LINKED ISSUE #427: Change drive selection UI to visually select relevant drives when All is chosen
This can make it easier to see what drives will be used for analysis.

In the current release let's say you have 1 drive selected:

> Image

And then you switch to **All Local Drives**. The list of drives remains unchanged despite selecting All. It's not *wrong*, but I found it visually confusing.

I suggest that when selecting **All Local Drives**, that the UI update to indicate exactly which drives will be used for analysis.

So, in this example, let's say I click **All Local Drives**, this is what you'd see:

> Image

And that way you know exactly what is going to happen when you go ahead and click **OK**.

Eilon:
Fixes #427


NoMoreFood:
Superseded

PR #431, opened 2026-02-24
Change Select Drives dialog layout

harryytm:
- moved OK and Cancel to lower right as Windows UX
- changed code to have them align to the right on resize
- using standard button size of 52x14
- changed radio buttons to take the whole width of the dialog
- changed bottom margin to 7 as top margin
image


windirstat/windirstat.rc
NoMoreFood: Could you also adjust these to use the full dialog width?

harryytm: On it

harryytm: amended as requested

Issue #333, opened 2025-11-28
Feature Request: Change/Limit Display Units (TB, GB, MB)


jbruchon:
When dir/file sizes cross a unit boundary, they switch to showing the higher unit. Unfortunately, especially for TB units, this can make size changes on refresh impossible to see. For example, I have a folder with 5.4 TB of data, deleted about 50 GB of data, and on refresh it shows I have 5.4 TB of data. I know that I can turn off the units entirely but that makes it difficult through added clutter. I'd like to be able to either:

- Change the visible units to always show as a certain unit i.e. everything is in MB, or
- Be able to set a unit "ceiling," i.e. anything crossing the TB boundary just shows thousands of GB instead

Neustradamus:
@NoMoreFood: Have you seen this @jbruchon ticket?

NoMoreFood:
@Neustradamus Seen it, yes, and considering it for a future release.

Forza-tng:
I'd prefer IEC units; KiB, MiB, GiB, etc. Roll over to next upper unit at 3 (999) or 4 (999.9) digits.

harryytm:
> I'd prefer IEC units; KiB, MiB, GiB, etc.

@Forza-tng WinDirStat already using IEC units by design, and corrected the size suffix units since 62e0363626c553f406da459289948ffbfcf107ff

> Roll over to next upper unit at 3 (999) or 4 (999.9) digits.

I am reverse engineering how Windows File Explorer does it just for fun, might be a good idea to do the same in WinDirStat.

Accorading my study, Windows File Explorer file property dialog would always show the file size in 3 significant digits at minimum, then move the decimal place around until exceeding 3 digit and the advance to next upper unit.

For example

- 9.99 MB - > 10.0 MB
- 99.0 MB -> 100 MB
- 999 MB -> 0.97 GB

Keep in mind the units in Windows are in fact IEC units, M$ just refused to use the correct display units for ages.

Additionally in detailed view, Windows File Explorer would always show size unit in KB as logic follows:
- <= 1 KB and > 0 byte as 1 KB
- 0 KB as 0 byte

harryytm:
@jbruchon just asking how would you like the controls to change the units?
I would envision the controls design like how 7-zip for changing date format in the menu, what do you think?

<img width="597" height="495" alt="Image" src="https://github.com/user-attachments/assets/551d3065-2d31-40a6-a54b-2f5e0c01db05" />

jbruchon:
[...]

PR #441, opened 2026-03-01
Minor Refactoring

harryytm:
- move Binary size constants to wds namesapce and reuse it in other part of the code
- correct units K M G T to Ki Mi Gi Ti

### DIFF
--- a/windirstat/Constants.h
+++ b/windirstat/Constants.h
@@ -32,6 +32,12 @@ namespace wds
     inline constexpr auto chrEqual        = L'=';
     inline constexpr auto szNpos          = std::wstring::npos;
 
+    // Binary size constants
+    inline constexpr auto Ki = 1024ull;
+    inline constexpr auto Mi = Ki * Ki;
+    inline constexpr auto Gi = Mi * Ki;
+    inline constexpr auto Ti = Gi * Ki;
+
     inline constexpr auto strExplorerKey = L"Software\\Microsoft\\Windows\\CurrentVersion\\Explorer";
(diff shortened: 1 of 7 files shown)


windirstat/Constants.h
```diff
@@ -32,6 +32,20 @@ namespace wds
...
     inline constexpr auto szNpos          = std::wstring::npos;
 
+    // Binary size constants
+    inline constexpr auto Ki = 1024ull;
+    inline constexpr auto Mi = Ki * Ki;
+    inline constexpr auto Gi = Mi * Ki;
+    inline constexpr auto Ti = Gi * Ki;
+
+    // Decimal size constants
+    inline constexpr auto K = 1000ull;
```
NoMoreFood: Where are these being used?

windirstat/Controls/FileDupeControl.h
```diff
@@ -52,9 +52,10 @@ class CFileDupeControl final : public CTreeListControl
 
     constexpr static ULONGLONG HashThresold(ITEMTYPE hashLevel)
     {
+        using namespace wds;
```
NoMoreFood: Let's use use wds::Ki and wds::Mi for these instead of using the namespace.

windirstat/Constants.h
harryytm: Had some ideas for #333 but I removed it in the amendment anyway.

windirstat/Controls/FileDupeControl.h
harryytm: Done in the amendment.

PR #445, opened 2026-03-05
SpaceMonger look in windirstat

v-pi:
Hello,

First, since I'm here, thanks a lot for your tool and the VERY simple workflow to start editing/building the project. That part (clone/build/run) was easier than expected.
I have always missed the look of SpaceMonger 1.4 and I have to admit I'm not a fan of the bubble look of WinDirStat. But this tool is so much better than SpaceMonger now, I wanted to try to add this feature myself.
Disclaimer, and I fear that's gonna be a deal-breaker but let's be honest, I have never written a single line of C++ in my life, and this is a fully vibe coded feature (credits to Gemini web and copilot Claude Haiku #notPayingForAnLLM).
As I'm typing this, copilot is still trying to complete these sentences (I'd tell him to shut up but he's right one time out of 20, and the rest of the time it's funny so, whatever. Also English is not my first language, so it's still a learning opportunity).
ANYWAYS...
What I wanted to achieve was: 
    - being able to display the borders without having the black hole of small files
    - adding a visible container for folders with the folder's name
Some screenshots attached. I've tried my best to make all the different options interact properly. Some decisions are intended (file names disappear when cushion shading is activated, shading is not applied to headers...). Most others are probably bugs.
This is intended to be entirely optional. Without checking "Show Headers" and with "Grid Area" at 0px, the rendering should be exactly the same as before. Emphasis on "should".
ANYWAYS, I'm quite happy with the result (the look at least, the code not so much but what am I going to do?) and I wanted to share this work in progress. This is a middle ground between a feature request and a pull request.
Still, this feels less lazy than just asking: "Hey, could you please add an entirely new look in WinDirStat?". But it's also clearly not some code you're going to merge as is.
I'm open to discussion, and I can even try to implement some of the feedback but again, not a C++ dev. I'm just going to feed your inputs into Gemini.

Some known issues (probably plenty more):
    - performance impact (I measured +30% render time for treemap T_T)
    - file/folder names are not readable when the background color is too dark (e.g., black)
    - still some hardcoded values so likely issues with DPI, large screens, etc.
    - editing colors in the options does not update the treemap until you re-render the treemap by another action
    - treeview preview does not display sample folder names
    - I have not checked the persistence/default values of the new options
    - The treemap option screen is cluttered now :(
    - I'm not sure the DrawTreeMap function respects all SOLID principles

2 new options: "Show Headers" and "Grid Minimum Area"
new_options

headers_flat

headers_with_cushion


NoMoreFood:
Thank you for the contribution.  I was considering _something_ similar.   I will consider merging this after the next release.

harryytm:
> headers_flat headers_with_cushion

@v-pi it looks super cool, I love it

v-pi:
Thanks for the kind words. I'm glad you like it ! I guess it means I have to finish it now :) I'll try to do my best to clean the code and polish the feature. Not sure about how this works after that though ? Once I am happy with my code you do an extensive code review ?

windirstat/Controls/TreeMap.cpp
```diff
@@ -427,55 +605,35 @@ CItem* CTreeMap::FindItemByPoint(CItem* item, const CPoint point)
...
 
-#ifdef _DEBUG
-            const CRect rcChild(child->TmiGetRectangle());
-            ASSERT(rcChild.right >= rcChild.left);
-            ASSERT(rcChild.bottom >= rcChild.top);
-            ASSERT(rcChild.left >= rc.left);
-            ASSERT(rcChild.right <= rc.right);
-            ASSERT(rcChild.top >= rc.top);
-            ASSERT(rcChild.bottom <= rc.bottom);
-#endif
```
harryytm: might be a good idea to put back debug block

v-pi:
OK, that took longer than expected... Would love to hear your thoughts on this new version. I tried my best to keep the original process of drawing borders by simply reducing the elements but the cost might be code readability. I'll let you judge. I eventually chose to discard the min size for grid rendering with a simple dropdown that adds an "Auto" value to the show grid parameter (in addition to the old show or don't show). I did not try to be backward compatible by handling a previously set "show grid" in the options, I hope that's ok.
If a refactoring of the drawtreemap is necessary, I'm probably not the best qualified person to do it but I can still try.

NoMoreFood:
I implemented something similar as an optional feature.  Thank you for the idea.

v-pi:
Thanks !

PR #452, opened 2026-03-13
Fix Extension Column Auto-size Bug

harryytm:
This fix resolves a defect where the auto-size functionality caused the Extension column to expand to an excessive width when items with no extension were present in the list.


harryytm:
@NoMoreFood although this fix the issue, the source of the problem still in `void CWdsListItem::DrawLabel`, I am figuring out how the code works and where to fix it, it should not report the width as 3503 if the text of the CWdsListItem is empty.

harryytm:
@NoMoreFood The latest force push 66e8c0374136fd149d72677342d262d555cf85a2 addressed the source of the problem in CWdsListItem::DrawLabel, and reverted the changes in CWdsListControl::GetSubItemWidth.

PR #453, opened 2026-03-13
Fix Extension Column Auto-size Bug Revision

harryytm:
Shift the fix even deeper to the source of the problem, which is cause by DrawTextCache::DrawTextCached aggressive return and done noting to the rect.


harryytm:
@NoMoreFood up to you to decide move the fix one level deeper or not, I really hope this is the bottom.

PR #459, opened 2026-03-18
Increase size and percentage display precision

harryytm:
- Increase size and percentage display precision to 2 decimal places
- Make string building buffer thread persistence
- Skip integer calculations for input lower than 0.005
- Fix a bug that UpdatePaneText would push MAXULONG64 directly into FormatBytes causing overflow input of 16777216

### DIFF
--- a/windirstat/HelpersInterface.cpp
+++ b/windirstat/HelpersInterface.cpp
@@ -130,15 +130,38 @@ std::wstring FormatCount(const ULONGLONG n) noexcept
     return FormatLongLongNormal(n);
 }
 
-std::wstring FormatDouble(const double d) noexcept
+std::wstring FormatDouble(const double input) noexcept
 {
-    ASSERT(d >= 0);
+    ASSERT(input >= 0);
+    static constexpr int scale = 100; // Scale to preserve two decimal places
+    thread_local std::wstring result; // Make string buffer thread persistent
+    result.clear();  // Clear previous contents while keeping allocated memory for reuse
 
-    const int x = std::lround(d * 10);
-    const int i = x / 10;
-    const int r = x % 10;
+    static constexpr auto appendInt = [](std::wstring& str, int n) noexcept {
+        const size_t start = str.size();
+        do { str.push_back(static_cast<wchar_t>(L'0' + (n % 10))); } while (n /= 10);
+        std::reverse(str.begin() + start, str.end());
+    };
 
-    return std::to_wstring(i) + GetLocaleDecimalSeparator() + std::to_wstring(r);
+    // Skip unnecessary integer calculations for very small values that returns 0.00
+    if (input < 0.005)
+    {
+        result.push_back(L'0');
+        result.push_back(GetLocaleDecimalSeparator());
+        result.append(L"00");
+        return result;
+    }
+
+    const int scaledValue = std::lround(input * scale);
+    const int integer = scaledValue / scale;
+    const int decimal = scaledValue % scale;
+
+    appendInt(result, integer);
+    result.push_back(GetLocaleDecimalSeparator());
+    if (decimal < 10) result.push_back(L'0'); // Pad with leading zero for single-digit decimals
+    appendInt(result, decimal);
+
+    return result;
 }
 
 std::wstring FormatFileTime(const FILETIME& t, const bool seconds) noexcept
--- a/windirstat/HelpersInterface.cpp
+++ b/windirstat/HelpersInterface.cpp
@@ -130,38 +130,16 @@ std::wstring FormatCount(const ULONGLONG n) noexcept
     return FormatLongLongNormal(n);
 }
 
-std::wstring FormatDouble(const double input) noexcept
+std::wstring FormatDouble(const double d) noexcept
 {
-    ASSERT(input >= 0);
-    static constexpr int scale = 100; // Scale to preserve two decimal places
-    thread_local std::wstring result; // Make string buffer thread persistent
-    result.clear();  // Clear previous contents while keeping allocated memory for reuse
+    ASSERT(d >= 0);
+    const int x = std::lround(d * 100);
+    const int i = x / 100;
+    const int r = x % 100;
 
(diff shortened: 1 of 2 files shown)


NoMoreFood:
Can you provide a better description for this and/or screenshots?

windirstat/HelpersInterface.cpp
```diff
@@ -130,15 +130,21 @@ std::wstring FormatCount(const ULONGLONG n) noexcept
...
+    const int scaledValue = std::lround(input * 100.0);
+    const int integer = scaledValue / 100;
+    const int decimal = scaledValue % 100;
 
-    const int x = std::lround(d * 10);
-    const int i = x / 10;
-    const int r = x % 10;
+    std::wstring result;
+    result.reserve(16); // Reserve enough space to avoid reallocations
+    result.append(std::to_wstring(integer));
```
NoMoreFood: This will all likely be less efficient than just doing std::wstring result = std::to_wstring(integer));

harryytm: I have used the debugger to find out `std::wstring result = std::to_wstring(integer));` would cause result being destroy and reinitiate a new result on every call.

By using a temporary VTRACE I found out FormatBytes would be called by multiple threads, hence using thread_local in the new amended commit.

Additionally I also found out `const auto sizeSummary = std::format(L"{}: \u2211 {}", Localization::Lookup(IDS_COL_SIZE_PHYSICAL), FormatBytes(size));;` in `MainFrame.cpp:1274` would directly pass `MAXULONG64` into `FormatBytes` causing an overflow input 16777216 and trigger the capacity increase from 7 to 15, which trigger a memory reallocation event.

harryytm: I had considered using std::format before instead of using `.append()` or `+=`, but due to `[[msvc::flatten]]` in the for loop of FormatSizeSuffixes, that would cause the building process stuck at linking stage and ever finish.

NoMoreFood: > I have used the debugger to find out `std::wstring result = std::to_wstring(integer));` would cause result being destroy and reinitiate a new result on every call.

The debugger is not a valid way to assess this.  With optimizations turned on, the second version will almost always be faster.

harryytm: > > I have used the debugger to find out `std::wstring result = std::to_wstring(integer));` would cause result being destroy and reinitiate a new result on every call.
> 
> The debugger is not a valid way to assess this. With optimizations turned on, the second version will almost always be faster.

Please clarify what does second version means?
Where do you want me to change?

NoMoreFood: Sorry, I meant:

std::wstring result;
result.reserve(16);
result.append(std::to_wstring(integer));

versus:

std::wstring result = std::to_wstring(integer);

harryytm: Does it mean you want me to remove
```cpp
    thread_local std::wstring result; // Make buffer thread persistent 
    result.clear(); // Clear previous content before appending new value
    result.append(std::to_wstring(integer));
```
and replace with
```cpp
std::wstring result = std::to_wstring(integer);
```

Additionaly with the bug resolved in UpdatePaneText, there would be almost zero chance the wstring cache `result` would increase capacity to 15 hence I removed 
```cpp
std::wstring result;
result.reserve(16);
```
in d75f6443f9ee04ced00ecd72dc82bba165ff5f3d and replaced with
```cpp
    thread_local std::wstring result; // Make buffer thread persistent 
    result.clear(); // Clear previous content before appending new value
```
in ca5d557aa65d6fe510514d47785d1462605e088e to make the cache presistance.

NoMoreFood:
I believe I implemented the functional effect of your change; let me know if you have any problems with it.  I didn't agree with some of the minor optimizations / guards you had in there.  Have to be careful with AI slop when it doesn't have the "bigger" picture and often doesn't take into account things like cache locality when adding tiny optimizations.

windirstat/MainFrame.cpp
```diff
@@ -1257,7 +1257,7 @@ void CMainFrame::UpdatePaneText()
 
     // Update select physical size
     const CClientDC dc(this);
-    const auto sizeSummary = (size != MAXULONG64) ? std::format(L"{}: \u2211 {}", Localization::Lookup(IDS_COL_SIZE_PHYSICAL), FormatBytes(size)) : wds::strEmpty;
```
harryytm, line 1260: @NoMoreFood This check is needed otherwise FormatBytes would be feed with MAXULONG64 and return 
Physical Size: ∑ 16777216 TiB to sizeSummary, wasting CPU for calculation that would never be used at the guard clause on line 1262.

harryytm: Image

harryytm:
> I believe I implemented the functional effect of your change; let me know if you have any problems with it.

there is a minor problem with your implementations, it did not implement to output .00 to make the number format visually uniformal as I intended, other than that the number is accurate and function, just me don't like how it looks.

> I didn't agree with some of the minor optimizations / guards you had in there. Have to be careful with AI slop when it doesn't have the "bigger" picture and often doesn't take into account things like cache locality when adding tiny optimizations.

Some of the optimizations was me prompting Gemini to do, and modify its suggestions, and me take "bigger" picture into account to decide what code it generated to use or not, for example it was me found out that there was more than one thread that would this function hence I decided to use thread_local instead of static, and if gard that skip calculation for input < 0.005 was my idea, not a suggestion from Gemini, including the naming of variables was hand-picked by me.

My mindset is to avoid any memory reallocations as possible, anything that can be reuse should be reuse instead of dispose of it and create a new one on every time the function is called, and with my limited knowledge, I believe that MSVC not yet smart enough to optimize those away, unless I have to learn to read the assembly code generated by the compiler to proof that, but at the moment I have some important thing to deal with first.

Anyway, thanks for your time reviewing the PR, it's been fun to make things better and I quite enjoy keep doing so, I hope my feedback does make sense for you.

PR #470, opened 2026-03-25
Improve TreeMap Zoom UX / Logic Refactoring

harryytm:
- retain parent / child selection and ensure zoom selection visible on TreeList during zoom
- refactor Zoom In and change canZoomIn to fit new UX logic
- fix zoom selection rectangle(blue frame) regression by force redraw TreeList during zoom
- fix reselect child would cause select all items on TreeList when selecting a child in a collapsed tree

### DIFF
--- a/windirstat/DirStatDoc.cpp
+++ b/windirstat/DirStatDoc.cpp
@@ -591,6 +591,10 @@ void CDirStatDoc::DeletePhysicalItems(const std::vector<CItem*>& items, const bo
 void CDirStatDoc::SetZoomItem(CItem* item)
 {
     m_zoomItem = item;
+    CTreeListControl* pControl = GetFocusControl();
+    pControl->EnsureItemVisible(item);
+    pControl->Invalidate();
+    pControl->UpdateWindow();
     UpdateAllViews(nullptr, HINT_ZOOMCHANGED);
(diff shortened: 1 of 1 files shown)


windirstat/DirStatDoc.cpp
```diff
@@ -591,6 +591,11 @@ void CDirStatDoc::DeletePhysicalItems(const std::vector<CItem*>& items, const bo
 void CDirStatDoc::SetZoomItem(CItem* item)
 {
     m_zoomItem = item;
+    [item](CTreeListControl* pControl) {
```
NoMoreFood: Using a lambda for a one-time operation like this hurts readability.  Just fetch the result from GetFocusControl() and deference that.

harryytm: will change that, but I have just found another bug during testing, canZoomIn logic not right

harryytm:
@NoMoreFood canZoomIn logic bug fixed, please review again, thanks

windirstat/DirStatDoc.cpp
harryytm: amended in the latest force-push 30bf047722a0923407d4e8316847bd11e605a677

PR #471, opened 2026-03-26
Fix Refresh Selected Crash

harryytm:
Fix a refresh selected bug which required the following items selection rules to avoid crash 

- IT_MYCOMPUTER can not mix with another item type
- IT_DRIVE, IT_DIRECTORY, IT_FILE can only mix with same type

### DIFF
--- a/windirstat/DirStatDoc.cpp
+++ b/windirstat/DirStatDoc.cpp
@@ -258,6 +258,23 @@ bool CDirStatDoc::HasRootItem() const
     return m_rootItem != nullptr;
 }
 
+// Validates whether the items selection would be safe to refresh
+bool CDirStatDoc::IsRefreshSelectionsSafe() const
+{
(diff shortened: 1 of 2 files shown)


NoMoreFood:
How did you manage to do a multiple selection with IT_MYCOMPUTER?

harryytm:
existing filters in OnUpdateCentralHandler allows it, as for doing that, holding Ctrl can multi select them in TreeList, press F5 after those selection would instant crash

harryytm:
image

NoMoreFood:
Gotcha.  I was looking at an older version.  This is actually a regression --- probably caused by the owner data change I made.  We should fix this in a different way. I will work on that.

NoMoreFood:
Should be addressed by: 
https://github.com/windirstat/windirstat/commit/6348f302adb8ea704a90b47f00fcd73d850ed2f5

PR #472, opened 2026-03-27
Fix fail to resolve custom folder icons issue

### LINKED ISSUE #354: Bug Report: Folder Custom Icon don't display properly on treelist
WinDirStat Version: Beta 2.4.4

Folders that have custom icon configured (via desktop.ini) does not display the custom icon properly, need to open context menu twice on treelist, and then collapse and expand the treelist to make the custom icon display correctly as workaround.

This bug existed for a long time but since it's not urgent I didn't report it.

harryytm:
This fix resolves #354

### DIFF
--- a/windirstat/Controls/IconHandler.cpp
+++ b/windirstat/Controls/IconHandler.cpp
@@ -62,6 +62,7 @@ void CIconHandler::Initialize()
         // Use two threads for asynchronous icon lookup
         m_lookupQueue.StartThreads(MAX_ICON_THREADS, [this]
         {
+            if (FAILED(CoInitializeEx(nullptr, COINIT_APARTMENTTHREADED | COINIT_DISABLE_OLE1DDE))) return;
             for (auto itemOpt = m_lookupQueue.Pop(); itemOpt.has_value(); itemOpt = m_lookupQueue.Pop())
             {
                 // Fetch item from queue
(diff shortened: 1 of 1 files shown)


windirstat/Controls/IconHandler.cpp
```diff
@@ -62,6 +62,7 @@ void CIconHandler::Initialize()
         // Use two threads for asynchronous icon lookup
         m_lookupQueue.StartThreads(MAX_ICON_THREADS, [this]
         {
+            CoInitializeEx(nullptr, COINIT_APARTMENTTHREADED | COINIT_DISABLE_OLE1DDE);
```
NoMoreFood: Should check return value (e.g., FAILED()) and just exit the thread if it fails (it should always succeed, but it's good practice).

harryytm: Might take time to figure out how to, I don't mind if you commit to this PR if it is quicker.

windirstat/Controls/IconHandler.cpp
```diff
@@ -62,6 +62,7 @@ void CIconHandler::Initialize()
         // Use two threads for asynchronous icon lookup
         m_lookupQueue.StartThreads(MAX_ICON_THREADS, [this]
         {
+            if (FAILED(CoInitializeEx(nullptr, COINIT_APARTMENTTHREADED | COINIT_DISABLE_OLE1DDE))) return;
```
harryytm, line 65: @NoMoreFood is this the check you talked about?

harryytm: the fix tested working in the test build

PR #474, opened 2026-03-27
Refined German translations

Minding000:
## Issues

- `IDS_MENU_EMPTY_FOLDER` said ~"This folder is empty" (state) instead of ~"I will empty this folder" (action)
- `IDS_CLEANUP_OPEN_IN_CONSOLE` and `IDS_MENU_CONSOLE` wording for "command prompt" is different from Windows' official wording
- `IDS_MENU_HELP_REPORT` is not an action (no verb)
- `IDS_USER_DEFINED_CLEANUP` referred to the first action explicitly
- `IDS_PAGE_TREEMAP_SCALE` is cut of in the UI for some reason

## Changes

- Fixed issues above
- Simplified wording in some places
- Capitalized "Bytes" (German noun)

## Screenshots

IDS_USER_DEFINED_CLEANUP
Screenshot 2026-03-27 140951
IDS_PAGE_TREEMAP_SCALE
Screenshot 2026-03-27 135618


harryytm:
Th label is 38x20 not going to be enough for the slider description, so I am prototyping something with Gemini, might solve the problem once and for all.
image

German testing screenshots
image

Minding000:
@harryytm I like the idea! I suppose that will be a follow up PR unrelated to this one, right?
Looking at the original `IDS_PAGE_TREEMAP_SCALE` screenshot, only one letter (and the following word) is cut of, so removing a letter (by changing "Skalierungs [...]" to "Skalierung") should fit for now or at least not make it worse.

harryytm:
I have submitted the PR, let see if it good enough to pass the review and get merged.

To me removing a letter just because the UI don't have enought space is unacceptable, with that change the label would have plenty of space vertically, just don't hurt your neck when you try to read though.

There are screenshot in that PR you can see how that would look like, I didn't even change IDS_PAGE_TREEMAP_SCALE

Minding000:
Cool
Just to be clear: "Skalierung" means "Scale", so the label is correct German with the removed letters (unlike the original broken label)

harryytm:
Give your thought on my PR, looks like NoMoreFood want to make the sliders horizontal instead ;(

PR #475, opened 2026-03-27
Add vertical Static Text support

harryytm:
- resolve text clipping issue in the text label of sliders in TreeMap page
- CJK aware to apply proper text scaling

Related to PR #474 

### Tested Language Screenshots
English
image
German
image
Japanese
image
Chinese Traditional
image
Korean
image


NoMoreFood:
I think it might look better just to rotate the sliders (left to right)

harryytm:
But I really love the vertical slider "Audio Mixer" cool look and want to keep it.
To me left to right sliders are boring.

Minding000:
This PR is definitely fixing the previous label display issues.
Despite the aesthetics, I agree with @NoMoreFood that horizontal sliders would be easier to read / more straight forward.
This is especially important with vision impaired users in mind.

harryytm:
like this?
image
To me I think this make WDS lost its signature look in this page. And for some reason the value label don't update in realtime when it is in horizontal mode, might need to change the code to make it work, but I really don't like the looks of this.

Minding000:
@harryytm Can you put the label above each slider (top-left) and the value in the top-right?
Then the text has a lot of space

harryytm:
this is the actual layout in VS2026
image
all those labels are above the sliders, the clipped part are caused by `\n` in the language files, and have to remove all it that in 23 languages

This is just a quick mod for me to proof that it look really boring and lack of the soul of WDS, so don't ask me to make this. And I don't think this is the right way to go.

PR #486, opened 2026-04-11
Add change comparison view for exported scan results

Asuta:
## Summary
- add a Changes tab and a `File -> Compare With Results...` flow to compare the current scan against a previously exported results CSV or snapshot CSV
- show positive size, file-count, and folder-count growth for matching paths, while keeping automatic history snapshot creation disabled
- optimize comparison and Changes view population so large result sets load faster, and preserve selection highlighting when clicking rows

## Notes
- includes a small `FinderNtfs.cpp` build fix needed for the current toolchain
- the new comparison flow is manual only; it does not write history files after each scan


windirstat/res/langs/lang_en.txt
```diff
@@ -7,6 +7,7 @@ IDS_ABOUT_TITLE=About WinDirStat
 IDS_ADMIN=Administrator
 IDS_ALL_FILES=All Files
 IDS_BACK_TO_SETTINGS=Back to\n&User Settings
+IDS_CHANGES=Changes
```
NoMoreFood, line 10: Could you use AI to produce the other language changes?  Here is the prompt I usually use which I then paste into PowerShell (pwsh.exe):

I am writing an application that scans a filesystem and displays information about file and directory statistics.  It also detects duplicate files by using hashes.  

Can you please translate the following strings from English to Czech (cs), Danish (da), German (de), Spanish (es), Estonian (et), Finnish (fi), French (fr), Hungarian (hu), Italian (it), Japanese (ja), Korean (ko), Dutch (nl), Norwegian (nb), Polish (pl), Portuguese (pt), Russian (ru), Slovenian (sl), Swedish (sv), Turkish (tr), Ukrainian (uk), Chinese – Hong Kong / Traditional (zh-HK), Chinese – Simplified (zh).  The translated string should be provided as lines of PowerShell that append the lines to a file named lang_xx.txt or lang_xx-YY.txt where 'xx' or 'xx-YY' is the two/four character code. Quote the string using double quotes.  Your response should be a single script that handles all languages.  <ADD IN A DESCRIPTION OF WHAT THESE STRINGS ARE MEANT TO DESCRIBE FOR BETTER ACCURARY>

IDS_<NAME>=<TEXT>

windirstat/Controls/TreeListControl.cpp
```diff
@@ -653,7 +653,7 @@ void CTreeListControl::ExpandItem(const CTreeListItem* item)
     ExpandItem(FindTreeItem(item), false);
 }
 
-void CTreeListControl::ExpandItem(const int i, const bool scroll)
+void CTreeListControl::ExpandItem(const int i, const bool scroll, const bool sort)
```
NoMoreFood: Why was this necessary?

windirstat/resource.h
```diff
@@ -248,7 +249,7 @@
 #ifdef APSTUDIO_INVOKED
 #ifndef APSTUDIO_READONLY_SYMBOLS
 #define _APS_NEXT_RESOURCE_VALUE        306
-#define _APS_NEXT_COMMAND_VALUE         32942
+#define _APS_NEXT_COMMAND_VALUE         32943
```
harryytm: Don't need to inclement the value of _APS_NEXT_COMMAND_VALUE if ID_COMPARE_RESULTS above was using and unused ID 32845.

harryytm:
Gemini 3 Pro PR review
---
## Comprehensive Architectural Review of the Scan History Compare Pull Request

The pull request under review, titled "Add scan history compare view," introduces a substantial amount of new functionality designed to allow users to compare directory snapshots and export the results to comma-separated values files. While the functional intent of the code is evident, a comprehensive examination of the implementation reveals several significant structural and architectural concerns. These concerns primarily revolve around how the new code integrates, or rather fails to integrate, with the established patterns of the existing software project. The following analysis details these issues in a structured and explicit manner to provide a clear understanding of the necessary modifications required before this code can be safely merged into the main project.

### Architectural Redundancy and the Duplication of Utilities

The most prominent issue within this pull request is the systematic recreation of utility functions that either already exist within the project or logically belong in the shared helper files. The author has implemented local, isolated functions for string encoding, specifically for converting text to and from the UTF-8 format. However, the software already possesses an established architecture for text manipulation within the shared interface helpers. By creating new, isolated versions of these functions, the author increases the overall size of the compiled program and introduces a redundant pathway for text processing. 

Similarly, the pull request includes localized logic for interacting with the file system, such as checking for the existence of directories and migrating legacy history files. The existing task helpers were explicitly designed to handle this exact type of low-level operating system interaction. Bypassing these established tools in favor of localized, hidden implementations creates a fragmented architecture where different parts of the application perform identical tasks using completely different underlying code. Furthermore, the introduction of custom time-formatting functions ignores the robust time-handling capabilities already present in the shared interface, forcing future developers to maintain multiple, disconnected standards for representing time across the application.

### The Ramifications of the Anonymous Namespace

A critical structural flaw in the proposed implementation is the extensive use of an anonymous namespace to encapsulate nearly four hundred lines of core logic. In the C++ programming language, an anonymous namespace is used to restrict the visibility of functions and variables strictly to the file in which they are defined. While encapsulation is generally a positive practice for hiding minor internal details, applying it to highly complex and fundamental logic, such as the custom comma-separated values parsing engine and the specialized hashing algorithms, creates a severely isolated environment. 

This isolation renders the encapsulated logic entirely untestable by automated unit testing frameworks. A developer cannot write an independent test to verify that the parsing engine correctly handles edge cases, such as embedded quotation marks or unusual characters, because the testing framework simply cannot access the function. Consequently, any future debugging efforts will require the developer to execute the entire snapshot comparison process rather than isolating the specific component that is failing. This lack of visibility transforms the file into an opaque container, significantly complicating long-term maintenance and verification.

### The Utilization of Unexplained Mathematical Constants

The pull request introduces specific, highly precise numeric literals to facilitate the hashing of file paths. Specifically, it employs the constants associated with the Fowler-Noll-Vo 1a hashing algorithm. However, the author has embedded these numbers directly into the source code without providing any explanatory comments or citations regarding their mathematical origin. The presence of such unexplained numbers, often referred to in software engineering as "magic numbers," poses a substantial risk to the stability of the system. 

If a future developer inadvertently modifies a single digit of these thirteen-digit constants during a routine code adjustment, the hashing algorithm will continue to function, but its mathematical distribution will be severely compromised. This degradation would lead to an increased rate of hash collisions, causing the comparison view to display inaccurate data. Without explicit documentation identifying these numbers as the Fowler-Noll-Vo constants, future maintainers will lack the necessary context to verify their accuracy or understand their critical role in the system's performance.

### Structural Overload and the Violation of Focused Responsibility

The structure of the newly introduced history management file attempts to handle too many disparate responsibilities simultaneously. A well-architected software component should ideally focus on a single, clearly defined task. In contrast, the proposed file acts concurrently as a file system manager, a data parser, a legacy format migrator, and an engine for calculating the growth differences between directory snapshots. 

This centralization of diverse tasks into a single location creates a highly coupled and brittle structure. If the project requirements change in the future, for example, if the software needs to export data to a different format instead of comma-separated values, the developer will be forced to modify this heavily overloaded file. This risks unintended disruptions to the unrelated file system management or data migration logic residing in the exact same location.

### Implications for Resource Management and Execution Speed

WinDirStat is fundamentally designed to be a highly responsive and lightweight analytical tool. However, the implementation patterns observed in this pull request threaten to degrade that performance standard. The code heavily relies on standard unordered map structures and performs continuous memory allocations for wide-character strings during the execution of the scan and compare processes. 

When applied to a file system containing millions of individual files, this approach will generate a massive volume of rapid memory allocations and deallocations in the system's memory manager. This phenomenon leads to memory fragmentation and heavily taxes the system resources, resulting in a noticeably sluggish user experience. Additionally, the inclusion of file migration logic deep within a function designed simply to retrieve a file path introduces the risk of unexpected blocking operations. If a user attempts to access the history view, the application might suddenly pause to physically move thousands of legacy files on the hard drive, causing the graphical user interface to freeze without any warning or indication of what the computer is doing.

### Final Recommendations for System Integrity

In its current state, the pull request introduces a high degree of technical liability. The isolated nature of the implementation, the duplication of existing utilities, and the lack of explicit documentation for critical mathematical components create a burdensome environment for future development. To ensure the long-term viability of the software, the implementation should be thoroughly restructured. The utility functions must be extracted and integrated into the established shared helper files. The critical parsing and hashing logic must be exposed for automated testing. Finally, the mathematical constants must be explicitly documented, and the resource-intensive migration logic should be separated into a distinct, predictable initialization phase rather than being hidden inside a simple retrieval function.

harryytm:
Gemini 3 Pro - PR risks and structural review
---
The pull request titled "Add scan history compare view" introduces several significant risks and structural issues. When evaluating the proposed changes against the existing architecture of the software, it becomes evident that the implementation deviates from established organizational rules and introduces potential points of failure. The following sections detail each specific problem related to the structure of the code and the risks it poses to the long-term maintenance and operation of the application.

### Duplication of Existing Utility Functions

A primary structural issue is the systematic recreation of utility functions that are already present within the shared helper files of the project. The author of the pull request has written localized functions for converting text encodings and formatting system time. However, the software already contains an established location for these exact operations. By writing new, isolated versions of these tools instead of utilizing the existing shared resources, the author increases the total size of the application unnecessarily. This duplication creates a situation where future programmers will have to maintain multiple, separate versions of identical logic, which increases the likelihood of inconsistencies and errors over time.

### Isolation of Logic Through Anonymous Namespaces

Another significant structural flaw is the extensive use of an anonymous namespace to conceal nearly four hundred lines of functional logic. While enclosing code within an anonymous namespace is a standard method for preventing naming conflicts, applying it to highly complex data parsing and hashing engines creates a severe testing limitation. Functions hidden in this manner cannot be accessed by automated testing frameworks. Consequently, a programmer cannot isolate and verify the correctness of the data parsing logic independently. If an error occurs in how the program reads a comma-separated values file, the programmer must execute the entire comparison process to locate the fault, making error detection and correction highly inefficient.

### Consolidation of Disparate Responsibilities

The newly introduced file is designed to perform too many unrelated tasks simultaneously. In a well-organized software system, a single file should have one primary responsibility. The proposed implementation, however, forces a single file to act as a system directory manager, a text data parser, a legacy file migrator, and an engine for calculating data differences. This high degree of consolidation means that the different parts of the logic are tightly connected. If a future update requires a change to the data export format, the programmer must modify a file that is also heavily involved in moving files on the hard drive. This increases the risk that a small change in one area will accidentally disrupt a completely different function.

### Risk of Undocumented Mathematical Constants

A major operational risk is the inclusion of highly specific numerical values for the hashing algorithm without any accompanying explanation. The code uses thirteen-digit constants required for the Fowler-Noll-Vo hashing calculation. Because there are no comments identifying these numbers or explaining their mathematical necessity, they are presented as arbitrary values. If a future programmer accidentally alters a single digit, the software will not generate an immediate error, but the effectiveness of the hashing process will degrade significantly. This degradation will cause the software to incorrectly identify different files as being identical. The lack of documentation forces future maintainers to guess the purpose of these critical numbers.

### Degradation of Hardware Resource Management

The proposed code introduces a high risk of degrading the overall execution speed and efficiency of the application. The logic relies heavily on continuous memory allocations and standard unordered map structures while processing the directory tree. When this logic is applied to a storage drive containing millions of individual files, the continuous request for and release of computer memory will cause significant memory fragmentation. This inefficient use of hardware resources will cause the application to run noticeably slower, which directly contradicts the primary goal of the software to act as a highly responsive storage analysis tool.

### Unpredictable Behavior from Hidden File Operations

Finally, there is a substantial risk of unpredictable application behavior due to the placement of file migration logic. The author has embedded the process for moving legacy history files deep within a function whose stated purpose is simply to retrieve a directory path. This means that when the application requests the location of the history folder, the program might suddenly begin moving thousands of files on the storage drive. This background operation will cause the main user interface to freeze without providing any visual warning or explanation to the user. Embedding active file operations inside simple retrieval functions makes the sequence of execution highly unpredictable and difficult to control.

harryytm:
Gemini 3 Thinking
---
## Analysis of Architectural Isolation and the Absence of Component Reusability

The implementation observed in this pull request presents a significant departure from the architectural standards established during the recent comprehensive reorganization of the codebase. A primary concern is the systematic isolation of core logic through the use of anonymous namespaces and localized utility functions. This approach effectively prevents any form of code reusability, which is a fundamental requirement for the ongoing preservation and careful management of the project’s technical integrity. By sequestering essential operations within a single translation unit, the author has created a structural environment that contradicts the principles of modular and efficient software development.

---

### The Consequence of Internal Linkage via Anonymous Namespaces

A substantial portion of the functional logic within this contribution is encapsulated within anonymous namespaces. In the C++ programming language, this specific design choice enforces internal linkage, which dictates that the included functions, structures, and variables are strictly invisible to every other component of the software. While this technique is sometimes appropriate for minor, file-specific utilities, its application here to core processes—such as directory hashing and string encoding—creates a significant barrier to the collaborative nature of the project.

Because these symbols are not exported to the linker, they cannot be accessed by other modules, even if the corresponding header file is included elsewhere. This results in the creation of a private, unreachable collection of logic. If another developer requires similar functionality for a different feature, they are unable to utilize the existing implementation. This necessitates either the duplication of code or a substantial refactoring of this entire module, both of which introduce unnecessary complexity and increase the likelihood of future software defects.

---

### Redundancy and the Proliferation of Duplicate Logic

The recent structural refinements in the project were specifically intended to centralize common operations into accessible interfaces, such as those found in the newly established helper files. This pull request, however, bypasses these centralized resources in favor of localized, redundant implementations. For example, the manual implementation of string conversion routines and path normalization logic within this PR represents a direct duplication of tools already provided by the project’s unified interface helpers.

This redundancy is problematic because it increases the overall volume of the source code without adding unique value. It creates what can be described as a shadow library—an unofficial set of tools that exists in parallel to the authorized project infrastructure. When a project maintains multiple versions of the same logic, it becomes significantly more difficult to ensure that optimizations or bug fixes are applied uniformly across the entire application. A refinement made to the primary helper file will not benefit this isolated module, leading to an inconsistent state where different parts of the software may exhibit different behaviors when performing identical tasks.

---

### Maintenance Burdens and the Fragmentation of the Codebase

The decision to avoid the established project-wide utilities in favor of isolated static helpers imposes a long-term burden on those responsible for the continued maintenance of the software. A unified codebase relies on the principle that a single responsibility should be handled by a single, authoritative component. By fragmenting the logic into separate, invisible pockets, this pull request complicates the process of auditing the code for performance and security.

Furthermore, this isolation prevents the implementation of modular testing. Because the core mathematical and system-level operations are trapped within the implementation file of the history module, they cannot be verified independently by external testing suites. This lack of visibility into the individual components of the feature makes it harder to identify the precise location of errors, thereby undermining the reliability of the system. The transition from a centralized, shared architecture back toward a collection of independent, non-reusable scripts represents a regression in the professional standards of the project.

harryytm:
@Asuta A list of existing WDS helpers you can use in your implementation, generated by Gemini 3 Thinking.

You can make use of these helpers once you included the common headers pch.h

---

### HelpersInterface.h / HelpersInterface.cpp

* **Name of the Signature**: `std::wstring GetLocaleString(LCTYPE lctype, LCID lcid)`
    * **Inputs**: `lctype` (The type of locale information requested); `lcid` (The identifier for the specific locale).
    * **Detail description**: This function interacts with the Windows operating system to retrieve specific strings related to regional settings. It is used to ensure that text-based information, such as names or measurements, conforms to the user's geographical and cultural preferences.
    * **Output**: A `std::wstring` containing the retrieved locale-specific text.

* **Name of the Signature**: `std::wstring GetLocaleLanguage(LANGID langid)`
    * **Inputs**: `langid` (The numerical identifier for a specific language).
    * **Detail description**: It queries the system database to find the human-readable name of a language associated with the provided ID. This allows the application to display language names in a clear and understandable manner within the interface.
    * **Output**: A `std::wstring` representing the full name of the language.

* **Name of the Signature**: `wchar_t GetLocaleThousandSeparator() noexcept`
    * **Inputs**: None.
    * **Detail description**: Retrieves the specific character used in the user's current locale to separate groups of thousands in large numbers. This is a critical component for ensuring that numerical data is displayed with proper regional formatting.
    * **Output**: A `wchar_t` representing the thousand separator character (e.g., a comma or a period).

* **Name of the Signature**: `wchar_t GetLocaleDecimalSeparator() noexcept`
    * **Inputs**: None.
    * **Detail description**: Obtains the character designated by the operating system to serve as the decimal point for the user's locale. It ensures that fractional numbers are rendered in a way that is familiar to the user.
    * **Output**: A `wchar_t` representing the decimal separator character.

* **Name of the Signature**: `std::wstring FormatBytes(ULONGLONG n) noexcept`
    * **Inputs**: `n` (The total count of bytes to be formatted).
    * **Detail description**: Converts a raw numerical byte value into a formatted string. It handles the logic for grouping digits and appending the correct binary size suffixes, such as "KiB" or "MiB," based on the user's configuration settings.
    * **Output**: A `std::wstring` containing the human-readable representation of the data size.

* **Name of the Signature**: `std::wstring FormatSizeSuffixes(ULONGLONG n) noexcept`
    * **Inputs**: `n` (The raw byte count).
    * **Detail description**: A specialized version of the size formatter that strictly applies binary size suffixes regardless of certain global settings. It ensures that the output is consistently presented with the appropriate magnitude indicator for technical clarity.
    * **Output**: A `std::wstring` with the formatted size and suffix.

* **Name of the Signature**: `std::wstring FormatCount(ULONGLONG n) noexcept`
    * **Inputs**: `n` (The numerical count of items).
    * **Detail description**: Takes a large integer and formats it as a string with the appropriate thousands separators. This is typically used for displaying counts of files or directories rather than data sizes.
    * **Output**: A `std::wstring` representing the formatted number.

* **Name of the Signature**: `std::wstring FormatDouble(double d) noexcept`
    * **Inputs**: `d` (The double-precision floating-point number to format).
    * **Detail description**: Converts a fractional number into a localized string. It ensures the decimal separator is correct and manages the number of decimal places to prevent visual clutter in the interface.
    * **Output**: A `std::wstring` containing the localized fractional number.

* **Name of the Signature**: `std::wstring FormatFileTime(const FILETIME& t, bool seconds = false) noexcept`
    * **Inputs**: `t` (A reference to the system file time structure); `seconds` (A boolean flag determining if seconds should be included in the output).
    * **Detail description**: Translates the Windows internal time representation into a legible date and time string. It respects the user's short-date and time formats as defined in the system's regional settings.
    * **Output**: A `std::wstring` representing the formatted date and time.

* **Name of the Signature**: `std::wstring FormatAttributes(DWORD attr) noexcept`
    * **Inputs**: `attr` (The bitmask containing file system attributes).
    * **Detail description**: Decodes the various flags associated with a file (such as Read-Only, Hidden, or System) and converts them into a concise string of characters. This provides a quick visual summary of the file's status.
    * **Output**: A `std::wstring` containing the attribute characters (e.g., "RHS").

* **Name of the Signature**: `std::wstring FormatPercentage(double d) noexcept`
    * **Inputs**: `d` (The decimal value to represent as a percentage).
    * **Detail description**: Multiplies a decimal value by one hundred and formats the result as a percentage string. It includes the localized decimal separator and the percentage symbol.
    * **Output**: A `std::wstring` representing the percentage.

* **Name of the Signature**: `std::wstring FormatCompressionRatio(double d) noexcept`
    * **Inputs**: `d` (The calculated compression ratio).
    * **Detail description**: Specifically formats the efficiency of file compression. It converts the ratio into a percentage string that clearly indicates the amount of space saved or used.
    * **Output**: A `std::wstring` containing the formatted compression ratio.

* **Name of the Signature**: `std::wstring FormatLongLong(ULONGLONG n) noexcept`
    * **Inputs**: `n` (The large integer value).
    * **Detail description**: Provides a generalized way to convert a 64-bit unsigned integer into a string with the correct thousand separators applied. It is a utility for displaying large values throughout the application.
    * **Output**: A `std::wstring` containing the formatted long integer.

* **Name of the Signature**: `std::wstring GlobToRegex(const std::wstring& glob, bool useAnchors = true)`
    * **Inputs**: `glob` (The string containing wildcard characters); `useAnchors` (A boolean indicating if the resulting pattern should be anchored to the start and end of a string).
    * **Detail description**: Performs a transformation of simple file system wildcards (like `*` or `?`) into complex regular expression syntax. This allows the application to use standard wildcard searches within advanced matching engines.
    * **Output**: A `std::wstring` containing the valid regular expression pattern.

* **Name of the Signature**: `std::wstring GetExtension(const std::wstring& path)`
    * **Inputs**: `path` (The full or partial file system path).
    * **Detail description**: Analyzes a path string to identify the file extension. It searches for the final period character and returns the subsequent characters, allowing the application to categorize files by type.
    * **Output**: A `std::wstring` containing the file extension.

* **Name of the Signature**: `std::vector<std::wstring> SplitString(const std::wstring& string, WCHAR delim = wds::chrPipe)`
    * **Inputs**: `string` (The source string to be divided); `delim` (The specific character used as a boundary for splitting).
    * **Detail description**: Segments a single string into multiple pieces based on the occurrence of the delimiter character. It also trims leading and trailing whitespace from each resulting segment to ensure clean data.
    * **Output**: A `std::vector` of wide strings representing the individual segments.

* **Name of the Signature**: `std::wstring GetAcceleratorString(UINT commandID)`
    * **Inputs**: `commandID` (The identifier for a specific menu or interface command).
    * **Detail description**: Retrieves the text representation of the keyboard shortcut (accelerator) associated with a given command. This is used to display shortcuts next to menu items for user convenience.
    * **Output**: A `std::wstring` containing the keyboard shortcut text.

* **Name of the Signature**: `bool IsControlKeyDown() noexcept`
    * **Inputs**: None.
    * **Detail description**: Queries the current state of the keyboard to determine if the "Control" key is being held down by the user. This is often used to modify the behavior of interface interactions.
    * **Output**: A boolean value indicating if the key is active.

* **Name of the Signature**: `bool IsShiftKeyDown() noexcept`
    * **Inputs**: None.
    * **Detail description**: Checks the current input state to see if the "Shift" key is depressed. It allows the application to provide alternative functionality during mouse or keyboard events.
    * **Output**: A boolean value indicating if the key is active.

* **Name of the Signature**: `void DrawTreeNodeConnector(CDC* pdc, const CRect& nodeRect, COLORREF bgColor, bool toTop, bool toBottom, bool toRight, bool showPlus = false, bool showMinus = false)`
    * **Inputs**: `pdc` (The device context for drawing); `nodeRect` (The boundaries of the node); `bgColor` (The background color); `toTop`, `toBottom`, `toRight` (Flags for line direction); `showPlus`, `showMinus` (Flags for expansion icons).
    * **Detail description**: A complex graphical utility that renders the connecting lines and expansion/collapse buttons used in the directory tree view. It ensures the visual hierarchy is clearly drawn on the screen.
    * **Output**: None (Performs direct graphical rendering).

---

### HelpersTasks.h / HelpersTasks.cpp

* **Name of the Signature**: `bool DriveExists(const std::wstring& path) noexcept`
    * **Inputs**: `path` (The path representing a disk drive).
    * **Detail description**: Verifies with the operating system whether a specified drive letter or volume is currently mounted and accessible for data retrieval. This prevents the application from attempting to scan non-existent hardware.
    * **Output**: A boolean indicating the existence and accessibility of the drive.

* **Name of the Signature**: `std::wstring GetVolumeName(const std::wstring& rootPath)`
    * **Inputs**: `rootPath` (The root directory of the volume, such as `C:\`).
    * **Detail description**: Requests the descriptive label or volume name assigned to a specific disk partition. This name is often chosen by the user or the system to identify the storage device.
    * **Output**: A `std::wstring` containing the volume label.

* **Name of the Signature**: `bool DeleteFileForce(const std::wstring& path, DWORD attributes = INVALID_FILE_ATTRIBUTES)`
    * **Inputs**: `path` (The path of the file to remove); `attributes` (Optional pre-queried file attributes).
    * **Detail description**: Attempts to delete a file using aggressive measures. If a standard deletion fails, it will attempt to remove restrictive attributes like "Read-Only" or "Hidden" before retrying, and may mark the file for deletion upon the next system restart.
    * **Output**: A boolean indicating if the deletion process was successfully initiated or completed.

* **Name of the Signature**: `std::wstring WdsQueryDosDevice(const std::wstring& drive)`
    * **Inputs**: `drive` (The drive letter to query).
    * **Detail description**: Resolves a standard Windows drive letter into its underlying internal system name (the MS-DOS device name). This is useful for identifying the physical hardware associated with a logical drive.
    * **Output**: A `std::wstring` representing the internal device name.

* **Name of the Signature**: `bool IsSUBSTedDrive(const std::wstring& drive)`
    * **Inputs**: `drive` (The drive letter to investigate).
    * **Detail description**: Checks if a specific drive letter is actually a virtual mapping created using the `subst` command, which links a drive letter to a folder path on another volume.
    * **Output**: A boolean indicating if the drive is a substituted path.

* **Name of the Signature**: `std::wstring GetDrive(const std::wstring_view& sv)`
    * **Inputs**: `sv` (A view of a path string).
    * **Detail description**: Extracts the drive letter component from a longer file system path. It typically isolates the first two characters (e.g., `C:`) to identify the source volume.
    * **Output**: A `std::wstring` containing the drive identifier.

* **Name of the Signature**: `void DisableHibernate() noexcept`
    * **Inputs**: None.
    * **Detail description**: Communicates with the system's power management service to temporarily prevent the computer from entering a hibernation state. This ensures that long-running directory scans are not interrupted by power-saving measures.
    * **Output**: None.

* **Name of the Signature**: `bool IsHibernateEnabled() noexcept`
    * **Inputs**: None.
    * **Detail description**: Queries the system settings to determine if the hibernation feature is currently active and supported by the hardware and operating system.
    * **Output**: A boolean indicating the hibernation status.

* **Name of the Signature**: `bool IsElevationActive() noexcept`
    * **Inputs**: None.
    * **Detail description**: Checks if the current instance of the application is running with administrative (elevated) privileges. This status determines whether the software can access restricted system areas.
    * **Output**: A boolean indicating if the process is elevated.

* **Name of the Signature**: `bool IsElevationAvailable() noexcept`
    * **Inputs**: None.
    * **Detail description**: Determines if the user's account has the capability to run the application with administrative privileges, even if it is not currently elevated.
    * **Output**: A boolean indicating if elevation is a possible option.

* **Name of the Signature**: `bool EnableReadPrivileges() noexcept`
    * **Inputs**: None.
    * **Detail description**: Attempts to adjust the security token of the application to include "Backup" and "Restore" rights. These special privileges allow the software to bypass certain security checks to scan system-restricted folders.
    * **Output**: A boolean indicating if the privileges were successfully acquired.

* **Name of the Signature**: `void RunElevated(const std::wstring& cmdLine)`
    * **Inputs**: `cmdLine` (The arguments to pass to the new process instance).
    * **Detail description**: Restarts the application in administrative mode. It ensures that all current configurations are saved before launching a new, elevated instance and closing the current one.
    * **Output**: None.

* **Name of the Signature**: `std::wstring GetNameFromSid(PSID sid)`
    * **Inputs**: `sid` (A pointer to a Windows Security Identifier).
    * **Detail description**: Translates a raw security identifier (SID) into a human-readable account name and domain. This is used to display the owner of a file or directory in a recognizable format.
    * **Output**: A `std::wstring` representing the account and domain name.

* **Name of the Signature**: `bool CompressFileAllowed(const std::wstring& volumeName, CompressionAlgorithm algorithm)`
    * **Inputs**: `volumeName` (The name of the target volume); `algorithm` (The specific compression method being checked).
    * **Detail description**: Queries the capabilities of a specific disk volume to see if it supports a particular compression technology, such as the standard NTFS compression or the newer modern formats.
    * **Output**: A boolean indicating if the algorithm is supported on the volume.

* **Name of the Signature**: `bool CompressFile(const std::wstring& filePath, CompressionAlgorithm algorithm)`
    * **Inputs**: `filePath` (The path to the file); `algorithm` (The compression method to apply).
    * **Detail description**: Commands the file system to apply a specific compression algorithm to a file. This reduces the physical space the file occupies on the disk without changing its logical content.
    * **Output**: A boolean indicating the success of the compression operation.

* **Name of the Signature**: `bool SparsifyFile(const std::wstring& path, ULONGLONG minZeroRunSize, ULONGLONG chunkSize)`
    * **Inputs**: `path` (The file path); `minZeroRunSize` (Minimum size of empty data to convert); `chunkSize` (The size of data blocks to process).
    * **Detail description**: Identifies large regions of empty data (zeros) within a file and instructs the file system to treat them as "sparse." This saves disk space by not storing the actual zeros while maintaining the file's original size.
    * **Output**: A boolean indicating if the file was successfully converted to a sparse format.

* **Name of the Signature**: `bool CreateHardlinkFromFile(const std::wstring& pathOne, const std::wstring& pathTwo)`
    * **Inputs**: `pathOne` (The existing file); `pathTwo` (The path for the new hard link).
    * **Detail description**: Creates a second directory entry that points to the exact same physical data on the disk as the original file. This allows one file to appear in two different locations without duplicating the storage space.
    * **Output**: A boolean indicating if the link was successfully created.

* **Name of the Signature**: `std::wstring ComputeFileHashes(const std::wstring& filePath)`
    * **Inputs**: `filePath` (The path of the file to be analyzed).
    * **Detail description**: Performs a mathematical analysis of a file's content to generate unique signatures or "hashes." These are used to identify files with identical contents even if their names are different.
    * **Output**: A `std::wstring` containing the calculated hash values.

* **Name of the Signature**: `void SetProcessPriority(DWORD priority)`
    * **Inputs**: `priority` (The desired system priority level).
    * **Detail description**: Adjusts how much CPU time the operating system allocates to the application. This can be used to lower the impact on the user's system or increase the speed of a scan.
    * **Output**: None.

* **Name of the Signature**: `void SetThreadPriority(HANDLE hThread, int priority)`
    * **Inputs**: `hThread` (The handle to a specific execution thread); `priority` (The priority level for that thread).
    * **Detail description**: Modifies the scheduling priority of a specific thread within the application, allowing more critical tasks to be completed ahead of less urgent ones.
    * **Output**: None.

* **Name of the Signature**: `void SetBackgroundProcessing(bool active)`
    * **Inputs**: `active` (A boolean toggle for background mode).
    * **Detail description**: Switches the application into a special processing mode that minimizes its use of system resources. This prevents the software from interfering with the user's other activities while a scan is in progress.
    * **Output**: None.

* **Name of the Signature**: `bool CreateShadowCopy(const std::wstring& volumePath)`
    * **Inputs**: `volumePath` (The path of the volume to snapshot).
    * **Detail description**: Uses the Volume Shadow Copy Service to create a temporary snapshot of a disk. This allows the application to read and scan files that are currently locked by other active programs.
    * **Output**: A boolean indicating if the shadow copy was successfully created.

* **Name of the Signature**: `void MapNetworkDrives()`
    * **Inputs**: None.
    * **Detail description**: Scans the system's registry to find disconnected network drive mappings and attempts to reconnect them. This ensures that all mapped network storage is available for analysis by the application.
    * **Output**: None.

windirstat/windirstat.vcxproj.filters
```diff
@@ -566,4 +584,4 @@
       <Filter>Resource Files\Languages</Filter>
     </Text>
   </ItemGroup>
-</Project>
\ No newline at end of file
+</Project>
```
harryytm: Please remove the new line at the end of this file

windirstat/windirstat.vcxproj
```diff
@@ -731,4 +737,4 @@
   </ImportGroup>
   <ImportGroup Label="ExtensionTargets">
   </ImportGroup>
-</Project>
\ No newline at end of file
+</Project>
```
harryytm: Please remove the new line at the end of this file

windirstat/res/langs/lang_en.txt
Asuta: Added the missing strings to the remaining language files, including the additional locale files that were still missing these entries.

windirstat/Controls/TreeListControl.cpp
Asuta: I reworked this to avoid changing the shared `CTreeListControl::ExpandItem` behavior. The Changes view now handles its own population locally, so the generic expansion path stays unchanged.

windirstat/resource.h
Asuta: Fixed. `_APS_NEXT_COMMAND_VALUE` is back to `32942` since `ID_COMPARE_RESULTS` uses the unused `32845` slot.

windirstat/windirstat.vcxproj.filters
Asuta: Fixed. Removed the trailing newline at the end of this file.

windirstat/windirstat.vcxproj
Asuta: Fixed. Removed the trailing newline at the end of this file.

harryytm:
Please amend 6bb2af6143c5e03a1bdfd9f46fd64c4f083dd92a commit title `复评审意见并补全多语言文案` and change it back to English, not everyone here able to read Chinese and it is out of place.

windirstat/Controls/TreeListControl.cpp
```diff
@@ -1,4 +1,4 @@
-﻿// WinDirStat - Directory Statistics
+// WinDirStat - Directory Statistics
```
harryytm: Revert this line change please

Asuta:
Updated the PR based on the latest feedback:

- restored the original BOM/header line in `TreeListControl.cpp`
- amended the last commit title to English

Thanks for the review.

windirstat/DirStatDoc.cpp
```diff
@@ -1139,6 +1155,48 @@ void CDirStatDoc::OnLoadResults()
...
+
+    if (resultsCsvCompare.status == ResultsCsvCompareStatus::InvalidResultsFile)
+    {
+        WdsMessageBox(L"The selected CSV is not a valid exported results file.", MB_OK | MB_ICONERROR);
+        return;
+    }
+
+    if (resultsCsvCompare.result.previousSnapshotLabel.empty())
+    {
+        WdsMessageBox(L"The selected CSV is not a supported results file.", MB_OK | MB_ICONERROR);
```
harryytm: Do not hardcode UI strings please.
Move them to language file and translate for all language or use existing IDS.

Asuta: Fixed. The compare-results CSV error messages now use localized IDS entries, with translations added to all language files.

NoMoreFood:
Closing due to unresolved comments.

PR #503, opened 2026-05-07
Minor Number Formatting Fixes

harryytm:
- Change usage percentage in volume name to use FormatDouble()
- ~~Add padding 00 to decimal places to form 123.00 or 0.00 instead of 123 or 0 to make the numbers look uniform in the column~~
- ~~Skip integer calculations that would resulted in 0.00~~

image

---

image

### DIFF
--- a/windirstat/HelpersInterface.cpp
+++ b/windirstat/HelpersInterface.cpp
@@ -133,11 +133,13 @@ std::wstring FormatCount(const ULONGLONG n) noexcept
 std::wstring FormatDouble(const double d) noexcept
 {
     ASSERT(d >= 0);
+    const std::wstring s = { GetLocaleDecimalSeparator(), L'0', L'0' };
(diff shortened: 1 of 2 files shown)


windirstat/HelpersInterface.cpp
```diff
@@ -133,11 +133,13 @@ std::wstring FormatCount(const ULONGLONG n) noexcept
 std::wstring FormatDouble(const double d) noexcept
 {
     ASSERT(d >= 0);
+    const std::wstring s = { GetLocaleDecimalSeparator(), L'0', L'0' };
```
harryytm: must not use static here to ensure always align with the setting of COptions::GetLocaleForFormatting()

harryytm:
@NoMoreFood Removed changes to FormatDouble which introduce excessive branching compared to latest verion.

PR #504, opened 2026-05-08
Add Settings Button To Toolbar

harryytm:
image


harryytm:
@NoMoreFood I am sorry  looks better than⚙, and ⚙looks like a ship's wheel rather than looks like a gear and I have tried that on before your edit, thus I chose  on purpose, please considered reverse to .


image

⚙
image

Besides, I have tried 🛠which is a much better choice, but your current implementation of `FromCharacter` don't accept it, `warning C4066: characters beyond first in wide-character constant ignored` would be shown during build.

Alternatively I have also tried ⚒ which does work with `FromCharacter` and it rendered as follows:
image

harryytm:
@NoMoreFood After digging through Windows Character Map, I found  or L'\xE15E', which look far more superior than another before and it works with existing `FromCharacter`
image

Render as follows
image

PR #505, opened 2026-05-08
Change configuration icon to a Wrench

harryytm:
@NoMoreFood After digging through Windows Character Map in Segoe UI Symbol, I found Wrench symbol  (L'\xE15E'), which is a far more superior choice than thick gear symbol  (L'xE115') and ⚒(L'x26CF'), due it is instantly understood by user that is for configuration(which is very important) and looks clean, it also works with existing `FromCharacter` without changing its code.
image

Render as follows on the toolbar

Small
image

Large
image


NoMoreFood:
I do not see this Segoe UI Symbol on Windows 7 / Server 2008 R2 (which I am trying to maintain compatibility with).

harryytm:
I installed a Windows 7 virtualbox and confirmed that, might be I try use the GDI+ approach.
That being said, ⚒ does exist in Windows 7's Segoe UI Symbol

harryytm:
Bug report: Pause and Help buttons has become square in Windows 7, further confirmed your concern beyond E115 I tried to use.
image

Issue #476, opened 2026-03-27
How to aggregate root-level files, as was previously the default?


codingatty:
I've used WinDirStat V1 for ages; currently running 1.1.2.80; now migrating to V2, 2.50 for the most excellent performance improvements.

V1 grouped all the root-level files into a group labeled \<Files>:

![Image](https://github.com/user-attachments/assets/b9739b02-1130-4690-adb1-1f31969613af)

V2 appears not to do that. It displays all individual files:

![Image](https://github.com/user-attachments/assets/cab4b1ef-f300-4b28-bfbe-85a067d2ee9a)

This makes it more difficult to do directory-to-directory comparisons. Is there any way to go back to the prior grouping? I cannot find anything in Options or Settings.

harryytm:
When you sort the name column, folder would come first then files.

codingatty:
> When you sort the name column, folder would come first then files.

I see, thanks. That's true; but then you're not sorting by size or date, making comparing directories more difficult.

Is it possible to have the files be grouped as before? Whether labelled "\<Files>" or "." or something else doesn't make much difference, but that was a very helpful feature, reducing noise when trying to look at directories comparatively.

harryytm:
Please describe your use case as clear and verbose as possible.

> When you sort the name column, folder would come first then files.

That was one of my Feature Request #259 which want implemented by @NoMoreFood in git commit f500c9acf7d1ab44c24edb0d8b69e8aa3476e4f1

harryytm:
Playing with the code in f500c9acf7d1ab44c24edb0d8b69e8aa3476e4f1 with the help of Google Gemin got me something like this, is that what you want?
<img width="1870" height="1221" alt="Image" src="https://github.com/user-attachments/assets/3f8ea048-f83d-48fb-8058-3527e4b82dea" />

<img width="1870" height="1221" alt="Image" src="https://github.com/user-attachments/assets/9ba345b8-8b7e-41a2-b597-df4bf9c871cd" />

harryytm:
It would be nice to allow user to enable/disable this kind of grouping or not, not all user prefer to have folders and files grouped.

harryytm:
@codingatty Just to report I have success on making the prototype work, with user selectable option in menu, and would update the sorting on setting change.

harryytm:
@codingatty here is a test build, please give feedback

[WinDirStat.zip](https://github.com/user-attachments/files/26355491/WinDirStat.zip)

[...]

PR #510, opened 2026-05-16
Restore ScrollToFirstItemByType

harryytm:
Scroll to behavior is the key feature and intention of ScrollToFirstItemByType and not a bug.

Original user requirement in #476

> [@harryytm](https://github.com/harryytm) , I've started using the test build you provided for all my uses of WinDirStat; and I've found that the "Group Folders Before Files" option actually groups the folders _after_ the files when sorting by last change (with most recent changes first). So when trying to examine the most recently changed folders,it's necessary to scroll past all the files first, after each refresh.
> 
> Is it possible to group the folders before files in all cases when "group folders before files" is selected?
> 
> [edit: clarify sorting with "most recent changes first"]

> > So when trying to examine the most recently changed folders,it's necessary to scroll past all the files first, after each refresh.
> 
> I am using Gemini to prototype a feature that scroll programmatically via a hotkey at the moment, it would select the first folder and scroll it to the top of the viewport as possible, might be that is what you want, this function don't need too much radical architectural changes to WDS, which might be easier to be merge into the main branch. You know I don't have the admin right of the repo, [@NoMoreFood](https://github.com/NoMoreFood) do.

### DIFF
--- a/windirstat/Controls/FileTreeControl.cpp
+++ b/windirstat/Controls/FileTreeControl.cpp
@@ -28,8 +28,8 @@ bool CFileTreeControl::GetAscendingDefault(const int column)
     return column == COL_NAME || column == COL_LAST_CHANGE;
 }
 
-// Select the first item of the same parent as the first selected item that matches any of the specified ITEMTYPE
-void CFileTreeControl::SelectFirstItemByType(ITEMTYPE itemType)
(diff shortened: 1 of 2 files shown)


NoMoreFood:
Can you see if the commit I just made addresses this?

harryytm:
Tested 5d1ce3b2c1a4e825bd5396d93c842ec3f9c3713c does addresses this, closing this PR.

PR #529, opened 2026-05-29
Refresh All On Filtering Settings Change

harryytm:


harryytm:
@NoMoreFood Might be a good idea to suppress the refresh if the filtering setting page is called from Select Drives, but I need time to read and understand the code to implement that, which involve some architectural changes, since you know the code better than me, would be easier and quicker if you can implement that using your knowledge on the code, for now this PR does its job enhancing the filtering UX.

PR #549, opened 2026-06-11
Feature: configurable I/O priority for scan and duplicate detection threads Patch 04/06

djdomi:
# Feature: Configurable I/O priority for scan and duplicate detection threads

To be honest with you before:

**What this means for you as a user:** WinDirStat now runs its disk reads at "Below Normal" I/O priority by default. Your foreground applications get first access to the disk; WinDirStat fills in the gaps. On an SSD the difference is barely noticeable. On an HDD or a busy system, it can make WinDirStat feel like it's running in the background where it belongs — scanning quietly without getting in your way. And if you actually *want* full speed (e.g. scanning an external drive with nothing else running), you can switch to Normal priority in Settings → Advanced.

## Summary

Adds a per-thread I/O priority setting that applies to all scan and duplicate-detection threads. Users can choose between Normal, Below Normal, and Idle I/O priority. The default is **Below Normal**, which keeps WinDirStat's disk access from competing with foreground applications during large scans.

## Background

Windows has a separate I/O scheduling layer from CPU scheduling. A thread running at `THREAD_PRIORITY_NORMAL` can still saturate disk I/O and cause noticeable stuttering in media players, build systems, and other I/O-heavy applications running concurrently. Setting thread-level I/O priority to `IoPriorityLow` (NT level 1) or `IoPriorityVeryLow` (NT level 0) tells the Windows I/O scheduler to service those I/O requests last, yielding to foreground traffic.

The NT thread I/O priority is set via `NtSetInformationThread` with `ThreadIoPriority` (class 31), which is undocumented but stable since Windows Vista and widely used.

## Files Changed

| File | Change |
|---|---|
| `windirstat/HelpersTasks.h` | Declared `SetCurrentThreadIoPriority(int level)` |
| `windirstat/HelpersTasks.cpp` | Implemented: maps 0/1/2 → NT values 2/1/0 |
| `windirstat/Options.h` | Added `static Setting<int> ScanIoPriority` |
| `windirstat/Options.cpp` | Registered setting with default 1 (Below Normal) |
| `windirstat/Pages/PageAdvanced.h` | Added `m_scanIoPriority` member |
| `windirstat/Pages/PageAdvanced.cpp` | Added combo box, `DDX`, persist on OK |
| `windirstat/DirStatDoc.cpp` | Call `SetCurrentThreadIoPriority` in scan thread lambda |
| `windirstat/Controls/FileDupeControl.cpp` | Call `SetCurrentThreadIoPriority` in `std::execution::par` lambda |

## Implementation Detail

The mapping between the user-facing levels and NT internals:

| User setting | NT `ThreadIoPriority` value | Meaning |
|---|---|---|
| 0 — Normal | 2 | Same as a foreground thread |
| 1 — Below Normal (default) | 1 | Low — serviced after normal I/O |
| 2 — Idle | 0 | Very Low — serviced last, only when disk is idle |

```cpp
void SetCurrentThreadIoPriority(const int level) noexcept
{
    constexpr ULONG ThreadIoPriority = 31;
    ULONG ioPriority = static_cast<ULONG>(2 - std::clamp(level, 0, 2));
    NtSetInformationThread(GetCurrentThread(),
        static_cast<THREADINFOCLASS>(ThreadIoPriority), &ioPriority, sizeof(ioPriority));
}
```

## Coverage

The setting is applied at every entry point where disk I/O occurs in the background:

1. **Scan threads** — set at `BlockingQueue::StartThreads` callback start in `StartScanningEngine`
2. **Dedup hash threads** — set inside the `std::execution::par` lambda in `RunDedupPhase`

The dedup phase uses `std::execution::par` (thread pool, not `BlockingQueue`), so `CancelThreadIo` doesn't reach those threads — but the I/O priority setting must be applied individually on each thread entry. Both are now covered.

## Settings UI

The combo box is added to the existing **Advanced** settings page, below the thread count selector:

- Normal
- Below normal *(default)*
- Idle


NoMoreFood:
The language files are not all localized properly.

djdomi:
_Originally posted by @NoMoreFood in https://github.com/windirstat/windirstat/issues/549#issuecomment-4680426742_

Hei @NoMoreFood 
Should be fixed as suggested by you. if it was not what you need, let me know about what was the intend you need

windirstat/resource.h
```diff
@@ -240,5 +240,6 @@
 #define _APS_NEXT_COMMAND_VALUE         32943
 #define _APS_NEXT_CONTROL_VALUE         1123
 #define _APS_NEXT_SYMED_VALUE           110
+#define IDC_SCAN_IO_PRIORITY            1134
```
NoMoreFood: This is in the wrong spot, has an outdated value, and _APS_NEXT_CONTROL_VALUE  will also need to incremented.

windirstat/Options.h
```diff
@@ -180,6 +180,8 @@ class COptions final
     static Setting<int> FollowReparsePointMask;
     static Setting<int> LanguageId;
     static Setting<int> FileHashAlgorithm;
+    static Setting<int>  ScanIoPriority;  // 0=Normal, 1=BelowNormal, 2=Idle
+
```
NoMoreFood: Remove entra line

windirstat/resource.h
harryytm: > This is in the wrong spot, has an outdated value, and _APS_NEXT_CONTROL_VALUE will also need to incremented.

@NoMoreFood As a Development oriented Feature Request, is it possible to have the build script handle _APS_NEXT values and possibly correcting placement and ID collision?

Time and time again PR that involve adding entries to resource.h was prone to ID collisions or _APS_NEXT not being updated. I think this can further reduce code maintenance overhead just as the current localizations workflow.

windirstat/HelpersTasks.cpp
```diff
@@ -752,6 +752,16 @@ void SetProcessIoPriorityHigh() noexcept
     }
 }
 
+void SetCurrentThreadIoPriority(const int level) noexcept
+{
+    // level: 0=Normal, 1=BelowNormal, 2=Idle
+    // NT ThreadIoPriority values: 2=Normal, 1=Low, 0=VeryLow
+    constexpr ULONG ThreadIoPriority = 31;
+    ULONG ioPriority = static_cast<ULONG>(2 - std::clamp(level, 0, 2));
```
harryytm: Not quite like the idea of using magic numbers, might be a good idea to use enum to make the values self documenting, instead of rely on other developer reading the comment to know what the numbers mean.

windirstat/Pages/PageAdvanced.cpp
```diff
@@ -131,6 +141,8 @@ void CPageAdvanced::OnOK()
     COptions::ProcessHardlinks = (FALSE != m_processHardlinks);
     COptions::ScanningThreads = m_scanningThreads + 1;
     COptions::FileHashAlgorithm = m_fileHashAlgorithm;
+    COptions::ScanIoPriority = m_scanIoPriority;
+
```
harryytm: Please remove extra line

windirstat/Pages/PageAdvanced.cpp
```diff
@@ -45,6 +45,8 @@ void CPageAdvanced::DoDataExchange(CDataExchange* pDX)
     DDX_Text(pDX, IDC_FOLDER_HISTORY_COUNT, m_folderHistoryCount);
     DDX_CBIndex(pDX, IDC_COMBO_THREADS, m_scanningThreads);
     DDX_CBIndex(pDX, IDC_HASH_ALGORITHM, m_fileHashAlgorithm);
+    DDX_CBIndex(pDX, IDC_SCAN_IO_PRIORITY, m_scanIoPriority);
+
```
harryytm: Please remove extra line

windirstat/Pages/PageAdvanced.h
```diff
@@ -53,6 +53,8 @@ class CPageAdvanced final : public CMFCPropertyPage
     BOOL m_processHardlinks = TRUE;
     int m_scanningThreads = 0;
     int m_fileHashAlgorithm = HASH_SHA512;
+    int m_scanIoPriority = 1;  // 0=Normal, 1=BelowNormal, 2=Idle
+
```
harryytm: Please remove extra line

windirstat/res/langs/lang_ja.txt
```diff
@@ -284,3 +284,4 @@ MSI_OptionsDlgDescription=以下のオプションを選択して、WinDirStat 
 MSI_OptionsDlgDesktopShortcut=デスクトップ ショートカットを作成する
 MSI_OptionsDlgHeaderTitle=インストール オプション
 MSI_OptionsDlgTitle=WinDirStat オプション
+IDS_PAGE_ADVANCED_IO_PRIORITY=スキャンI/O優先度
```
harryytm: Missing padding spaces surrounding `I/O` which violates CJK typography guidelines, suggest correcting as follows:

Before:
`スキャンI/O優先度`

After:
`スキャン I/O 優先度`

windirstat/Pages/PageAdvanced.cpp
djdomi: Fixed in PR

windirstat/HelpersTasks.cpp
djdomi: Fixed in PR-Update

windirstat/Options.h
djdomi: Fixed in PR

windirstat/resource.h
djdomi: im not sure what i can do here, i think i fixed it and no conflicts are currently open?

windirstat/Options.h
```diff
@@ -186,6 +186,8 @@ class COptions final
...
```
harryytm: Please remove extra line

windirstat/Pages/PageAdvanced.cpp
```diff
@@ -95,6 +96,14 @@ BOOL CPageAdvanced::OnInitDialog()
...
     m_scanningThreads = COptions::ScanningThreads - 1;
     m_fileHashAlgorithm = COptions::FileHashAlgorithm;
+    m_scanIoPriority = COptions::ScanIoPriority;
+    // Populate scan I/O priority combo
+    if (auto* pIoPrio = static_cast<CComboBox*>(GetDlgItem(IDC_SCAN_IO_PRIORITY)); pIoPrio != nullptr)
+    {
+        static constexpr std::wstring_view kLabels[] = { L"Normal", L"Below normal", L"Idle" };
+        for (const auto& label : kLabels)
+            pIoPrio->AddString(label.data());
+    }
```
harryytm: I think there is 2 issues with this block

1. `L"Normal", L"Below normal", L"Idle"` should not be hardcoded strings, please create IDS in localizations files for each of them

2. Placement of this if-block should be placed after `m_folderHistoryCount = std::to_wstring(COptions::FolderHistoryCount.Obj()).c_str();` and before `UpdateData(FALSE);`, and should a empty lines surrounding it to improve code readability.

windirstat/Options.h
harryytm: Please remove extra line and excess whitespace before `ScanIoPriority`

Example: `static Setting<int> ScanIoPriority;  // 0=Normal, 1=BelowNormal, 2=Idle`

harryytm:
I have noticed the language files are not sorted, please build the solution(in Visual Studio) to trigger the project built-in build script do that for you before commiting.

djdomi:
> I have noticed the language files are not sorted, please build the solution(in Visual Studio) to trigger the project built-in build script do that for you before commiting.

Upsi, taken a shortcut :D fixing it and respecting the way how this project works.

harryytm:
> > I have noticed the language files are not sorted, please build the solution(in Visual Studio) to trigger the project built-in build script do that for you before commiting.
> 
> Upsi, taken a shortcut :D fixing it and respecting the way how this project works.

Let me explain, WinDirStat project solution have a built-in pre-build script that would run on `Build Solution` in Visual Studio, it would sort all the new IDS for you and add required code to LangStrings.h for you automatically.

You don't need to sort the lines in lang_???.txt or adding code to LangStrings.h, all you need to do is build the solution.

djdomi:
> > > I have noticed the language files are not sorted, please build the solution(in Visual Studio) to trigger the project built-in build script do that for you before commiting.
> > 
> > 
> > Upsi, taken a shortcut :D fixing it and respecting the way how this project works.
> 
> Let me explain, WinDirStat project solution have a built-in pre-build script that would run on `Build Solution` in Visual Studio, it would sort all the new IDS for you and add required code to LangStrings.h for you automatically.

Yeh i agree, i have spoken with my "agent" to respect this situation, we are workin on using the vcxproj file :)

djdomi:
@harryytm i addressed your words as also @NoMoreFood i hope :)

harryytm:
I am not part of the WinDirStat team, I just advice you what might need to change, now is down to him to decide what else need to change.

For me I still have a lot of unmerged PR hanging around. ;(

windirstat/FinderNtfs.cpp
```diff
@@ -210,7 +210,7 @@ bool FinderNtfsContext::LoadRoot(CItem* driveitem)
     // Process MFT records
     std::for_each(std::execution::par, dataRuns.begin(), dataRuns.end(), [&](const auto& dataRun)
     {
-        // Page alignment satisfies FILE_FLAG_NO_BUFFERING for any sector size 
+        // Page alignment satisfies FILE_FLAG_NO_BUFFERING for any sector size
```
harryytm: Remove these changes which don't belongs to your PR to reduce diff noise.
Created PR #556 to address it, which have bothered me for quite some time.

djdomi:
Removed the unrelated `FinderNtfs.cpp` whitespace change as requested — it was a stray diff from the working copy base, not intentional. Thanks for catching it @harryytm!

PR #550, opened 2026-06-11
Fix: stop/cancel scan hangs when ReadFile blocks in worker thread 01/06

djdomi:
# Fix: WinDirStat hangs at 99% / "Cancel" button stops responding

To be honest with you before:
(i pre-wrote the texts and github drives me crazy for getting the right text to the right patch :D)

This one drove me crazy. You start a scan, decide you want to stop it — maybe you picked the wrong drive, maybe it's taking too long — and you click Cancel. 
Nothing happens. The window locks up. The progress bar freezes at 99%. 
The only way out is to open Task Manager and kill the process. 
You lose whatever scan data you had, and on top of that you feel like the application let you down.
I am not sure if this always happends to others, but i share my findings what I worked through this problem together with Claude (Anthropic's AI assistant), 
which helped me understand what was actually happening inside the code, trace the root cause, and write a proper fix. 

**What this means for you as a user:** After this patch, clicking Stop or closing the window during a scan responds immediately — even when WinDirStat is in the middle of reading a large file on a slow HDD. No more killing the process. No more frozen windows. The application just... stops, cleanly, the way it should have always worked.

## Symptom

After clicking **Stop** or closing WinDirStat mid-scan, the application freezes. The progress bar (may) stay(s) at 99%, 
the window becomes unresponsive, and Task Manager shows CPU at 0% — the process is stuck, not working. The only way out is to kill the process.

## Root Cause

`BlockingQueue::SuspendExecution()` waits in a loop for `AllThreadsIdling()` to return true. This condition becomes true only when every worker thread is blocked inside `Pop()`, waiting for the next work item.

The problem: a worker thread reading a large file is inside `ReadFile()` — not inside `Pop()`. It is actively doing I/O, not waiting. So `AllThreadsIdling()` never becomes true, and `SuspendExecution()` waits forever.

```
UI thread:  StopScanningEngine()
              → queue.SuspendExecution()          ← blocks here waiting for AllThreadsIdling()
                  → m_waiting.wait(lock, [&] { return AllThreadsIdling(); })
                                                   NEVER returns ↑

Worker thread: ReadFile(hFile, …)                 ← blocked here, not in Pop()
```

`CancelSynchronousIo(HANDLE)` is the Windows API designed exactly for this: it sends a cancellation signal to a specific thread's pending synchronous I/O operation. `ReadFile` returns immediately with `ERROR_OPERATION_ABORTED`, the worker exits the I/O call, returns to its `Pop()` loop, and can now reach the idle state.

## Files Changed

| File | Change |
|---|---|
| `windirstat/BlockingQueue.h` | Added `CancelThreadIo()` method |
| `windirstat/DirStatDoc.cpp` | Call `CancelThreadIo()` on all queues before `SuspendExecution` in `StopScanningEngine` |

## The Fix

**`BlockingQueue.h`** — new method:
```cpp
void CancelThreadIo()
{
    std::scoped_lock lock(m_mutex);
    for (auto& thread : m_threads)
    {
        if (thread.joinable())
            CancelSynchronousIo(thread.native_handle());
    }
}
```

**`DirStatDoc.cpp`** — called before every `SuspendExecution`:
```cpp
void CDirStatDoc::StopScanningEngine(StopReason stopReason)
{
    // Interrupt blocking I/O so workers reach WaitIfSuspended promptly.
    for (auto& queue : m_queues | std::views::values)
        queue.CancelThreadIo();

    // Now safe to suspend — workers will reach idle state quickly.
    for (auto& queue : m_queues | std::views::values)
        ProcessMessagesUntilSignaled([&queue] { queue.SuspendExecution(); });
    …
```

## Why This Is Correct and Not a Band-Aid

- `CancelSynchronousIo` is documented in MSDN for exactly this purpose: "cancels a pending synchronous I/O operation that is issued by the calling thread"
- The ReadFile call returns `ERROR_OPERATION_ABORTED` — the worker handles this gracefully, closes the file handle, and moves on
- The root cause (deadlock between `AllThreadsIdling` and in-flight I/O) is eliminated, not masked
- The existing scan suspend path (user-initiated pause) intentionally does **not** use `CancelThreadIo` — a paused scan should finish reading the current cluster, which is the correct behavior

## Tested Scenarios

- Stop scan while reading a large ISO/VHD on a slow HDD → stops within ~50 ms
- Stop scan on a fast SSD → no visible change (was already fast)
- Close application mid-scan → exits cleanly without Task Manager required
- Suspend/Resume scan → unaffected (does not call `CancelThreadIo`)

### DIFF
--- a/windirstat/BlockingQueue.h
+++ b/windirstat/BlockingQueue.h
@@ -175,9 +175,6 @@ public:
 
     void CancelThreadIo()
     {
-        // Interrupt any blocking I/O (ReadFile, etc.) in worker threads so they
-        // can reach WaitIfSuspended quickly instead of waiting for the read to
-        // complete naturally. ReadFile returns ERROR_OPERATION_ABORTED on cancel.
         std::scoped_lock lock(m_mutex);
         for (auto& thread : m_threads)
         {
--- a/windirstat/BlockingQueue.h
+++ b/windirstat/BlockingQueue.h
@@ -173,6 +173,19 @@ public:
         return m_stopReason;
     }
 
+    void CancelThreadIo()
+    {
+        // Interrupt any blocking I/O (ReadFile, etc.) in worker threads so they
+        // can reach WaitIfSuspended quickly instead of waiting for the read to
+        // complete naturally. ReadFile returns ERROR_OPERATION_ABORTED on cancel.
+        std::scoped_lock lock(m_mutex);
+        for (auto& thread : m_threads)
+        {
+            if (thread.joinable())
+                CancelSynchronousIo(thread.native_handle());
+        }
+    }
+
     void CancelExecution(const int stopReason = -1)
     {
         // Start cancellation process
--- a/windirstat/DirStatDoc.cpp
+++ b/windirstat/DirStatDoc.cpp
@@ -1740,6 +1740,13 @@ void CDirStatDoc::OnScanStop()
 
 void CDirStatDoc::StopScanningEngine(StopReason stopReason)
 {
+    // Interrupt blocking I/O (e.g. ReadFile on large files) in worker threads
+    // so they reach WaitIfSuspended promptly. Without this, SuspendExecution
+    // hangs indefinitely waiting for AllThreadsIdling() while a thread reads.
+    for (auto& queue : m_queues | std::views::values)
+        queue.CancelThreadIo();
+
+
     // Request for all threads to stop processing
     for (auto& queue : m_queues | std::views::values)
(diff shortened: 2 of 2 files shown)


NoMoreFood:
There's an extra newline in StopScanningEngine() and please remove the duplicate comment in CancelThreadIo()

djdomi:
> There's an extra newline in StopScanningEngine() and please remove the duplicate comment in CancelThreadIo()

Hei @NoMoreFood 
As Also here, should be fixed *hopefully* :)

PR #552, opened 2026-06-11
Fix: right-click context menu freezes UI when many files are selected 03/06

djdomi:
# Fix: Right-click context menu freezes UI when many files are selected

To be honest with you before:
I'm not really a programmer. 
I'm just a regular WinDirStat user who got frustrated enough with this specific bug to actually do something about it.

The freeze is/was real and repeatable: 
select a bunch of large files in the tree, right-click, 
and WinDirStat locks up completely for several seconds before the menu appears. 
On some of my machines it took more as 10+ seconds. 
During that time you can't even move the window. 
It felt like the application had crashed. 
If you're in a hurry which you usually are when you're cleaning up disk space this is genuinely painful.

I tracked down the cause together with Claude (Anthropic's AI), 
which helped me understand how Windows shell extensions work and why they make WinDirStat block. 
I want to be honest: this fix was developed with AI assistance. 
But it solves the real problem, it was validated against the actual code, and I've tested it on my own machines.

**What this means for you as a user:** Right-clicking even large multi-file selections now feels instant. If shell extensions still need a moment to load (OneDrive is especially slow), you at least see a wait cursor instead of a frozen window. Whether you're on a corporate machine with a dozen shell extensions or a clean personal PC, the context menu is now responsive.

## Symptom

Selecting 50+ files in the file tree and right-clicking causes WinDirStat to freeze for several seconds (sometimes 10–30+ seconds) before the context menu appears. On systems with OneDrive, antivirus software, or git installed, the freeze is especially long. During this time the window is completely unresponsive.

## Root Cause

The Windows shell context menu system (`IContextMenu`) loads shell extensions synchronously on the calling thread. When you pass a list of file paths to `GetContextMenu()` and `QueryContextMenu()`, every registered shell extension (OneDrive sync overlay, Windows Defender, Git integration, etc.) is initialized for each path in the list. This loading happens on the UI thread, blocking it.

With 100 selected files, a shell extension that takes 50 ms per file adds 5 seconds of UI freeze. Extensions that make network calls (OneDrive status check, antivirus cloud lookup) can take much longer.

This affected two call sites independently:
1. `CMainFrame::OnInitMenuPopup` — right-click in the main file tree
2. `CDirStatDoc::OnContextMenuExplore` — "Open in Explorer" context action

Both sites were missing a path cap and provided no visual feedback while loading.

This might be a bit more dramastic told as is but sometimes are some seconds, that were the thing you dont need when been in hurry :)
## Files Changed

| File | Change |
|---|---|
| `windirstat/HelpersInterface.h` | Added `constexpr kShellMenuPathLimit = 15` (shared constant) |
| `windirstat/MainFrame.cpp` | Cap paths + `BeginWaitCursor`/`EndWaitCursor` around both shell calls |
| `windirstat/DirStatDoc.cpp` | Same cap + wait cursors at the second call site |

## The Fix

**Centralized constant** (one place to tune):
```cpp
// HelpersInterface.h
constexpr size_t kShellMenuPathLimit = 15;
```

**Applied identically at both call sites:**
```cpp
if (paths.size() > kShellMenuPathLimit)
    paths.resize(kShellMenuPathLimit);

BeginWaitCursor();
const CComPtr contextMenu = GetContextMenu(hwnd, paths);
EndWaitCursor();

if (contextMenu != nullptr)
{
    BeginWaitCursor();
    contextMenu->QueryContextMenu(…);
    EndWaitCursor();
}
```

The limit of 15 is chosen because:
- Shell extensions scale linearly with path count; 15 paths is unnoticeable on any system
- Beyond ~20 paths, `IShellFolder::GetUIObjectOf` may reject the list anyway if paths span different parent directories
- The menu items produced for a 15-path selection are identical in practice to those for a 200-path selection (Cut, Copy, Delete, Properties — all work on whatever is selected)

The `BeginWaitCursor` / `EndWaitCursor` pair provides visual feedback on the cases where the cap doesn't help (i.e. the first 15 paths include slow extensions).

## Idempotency

Both call sites were independently fixed with the same constant and the same pattern. There is no third call site for context menus in this codebase (`grep GetContextMenu` confirms exactly two).

### DIFF
--- a/windirstat/DirStatDoc.cpp
+++ b/windirstat/DirStatDoc.cpp
@@ -1767,16 +1767,23 @@ void CDirStatDoc::OnContextMenuExplore(UINT nID)
     for (const auto& item : selected)
         paths.emplace_back(item->GetPath());
 
-    // query current context menu
     if (paths.empty()) return;
+    if (paths.size() > kShellMenuPathLimit)
+        paths.resize(kShellMenuPathLimit);
+
+    AfxGetMainWnd()->BeginWaitCursor();
     const CComPtr contextMenu = GetContextMenu(CMainFrame::Get()->GetSafeHwnd(), paths);
+    AfxGetMainWnd()->EndWaitCursor();
     if (contextMenu == nullptr) return;
 
     // create placeholder menu
     CMenu menu;
     if (menu.CreatePopupMenu() == 0) return;
-    if (FAILED(contextMenu->QueryContextMenu(menu.GetSafeHmenu(), 0,
-        CONTENT_MENU_MINCMD, CONTENT_MENU_MAXCMD, CMF_NORMAL))) return;
+    AfxGetMainWnd()->BeginWaitCursor();
+    const bool queryOk = SUCCEEDED(contextMenu->QueryContextMenu(menu.GetSafeHmenu(), 0,
+        CONTENT_MENU_MINCMD, CONTENT_MENU_MAXCMD, CMF_NORMAL));
+    AfxGetMainWnd()->EndWaitCursor();
+    if (!queryOk) return;
 
     // launch command associated with passed item identifier
     CMINVOKECOMMANDINFOEX info = {};
--- a/windirstat/HelpersInterface.h
+++ b/windirstat/HelpersInterface.h
@@ -141,3 +141,5 @@ inline CRect ClientRectOf(const CWnd* hWnd)
     ::GetClientRect(*hWnd, &rc);
     return rc;
 }
+
+constexpr size_t kShellMenuPathLimit = 15;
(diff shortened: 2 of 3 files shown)


NoMoreFood:
Not interested in this one.  The wait cursor is fine (although we have an existing RAII class for this).  There is protective right-click logic that prevent menu certain functions from being executed on certain paths.   It's likely that one of the checks is slow on your system and we should figure out which check is slow and how we can accelerate that.

djdomi:
i am fine, when you dont like it, i am just give it as a gift 👍 
Feel free to close or improve it i just give back what i have modified on me to be align with OSS and share my modifikations.

PR #554, opened 2026-06-11
Feat/Fix: GPU benchmark worker thread data race on HWND + file load on UI - 06/06

djdomi:
# Fix: GPU benchmark worker thread data race on HWND + UI freeze during file load

To be honest with you before:
(yup last one: i repeat at this point already multiple times myself, because i like to be honest with you :D)

This fix came out of a broader code review I was doing (with Claude's help) of all the threading and UI interaction patterns in WinDirStat's newer features.
The GPU/CPU hashing benchmark is a great addition to the application — it helps users pick the fastest hash algorithm for their hardware. 
But the implementation had two issues that could cause problems: 
the file loading happened on the UI thread (freezing the window while reading a large test file), 
and after that was moved to the worker thread, `GetSafeHwnd()` was still being called from the worker which is a threading violation in MFC. 
Thread sanitizers would flag it, and under the right race conditions it could cause a crash.

The analysis is correct and the fix is clean: 
the window handle is captured once, on the UI thread, 
before the worker starts the way it should have been done from the beginning.

**What this means for you as a user:** Clicking "Run Benchmark" with a large test file no longer freezes the settings window while the file loads. 
The benchmark runs properly in the background, the UI stays responsive throughout, and closing the settings dialog while a benchmark is running is safe.

## Symptom

Two issues in the GPU/CPU hashing benchmark tab:

1. **File load blocks UI** — Clicking "Run Benchmark" with a large file (e.g. a multi-GB ISO) caused WinDirStat to freeze while reading the test data, because the file I/O happened on the UI thread.

2. **Data race on `m_hWnd`** — After the fix moved file loading to a worker thread, `GetSafeHwnd()` was still called from inside the worker thread. `GetSafeHwnd()` reads `CWnd::m_hWnd`, which is written on the UI thread (set to NULL when the window is destroyed). Reading and writing `m_hWnd` from two different threads without synchronization is a data race — technically undefined behavior, flagged by thread sanitizers.

## Root Cause

Original code (file load on UI thread, then thread starts):
```cpp
// UI thread — blocking file read before spawning thread
std::vector<BYTE> testData = …ReadFile(m_testFilePath)…;
m_benchThread = std::jthread([this, data = std::move(testData)](…) {
    RunBenchmark(stopToken, std::move(data));
});
```
Moved to worker but still calling `GetSafeHwnd()` from it:
```cpp
void CPageGpu::RunBenchmark(…) {
    const HWND hwnd = GetSafeHwnd();  // ← worker thread reading m_hWnd
```

And the early-exit error path in the lambda also called `GetSafeHwnd()` directly:
```cpp
::PostMessage(GetSafeHwnd(), WM_BENCH_DONE, 1, 0);  // ← same issue
```

## Files Changed

| File | Change |
|---|---|
| `windirstat/Pages/PageGpu.cpp` | Move file load into worker; pre-capture HWND before spawning thread; add `THREAD_PRIORITY_BELOW_NORMAL`; add `SleepEx` yield between algo measurements; handle error flag in `OnBenchmarkDone` |
| `windirstat/Pages/PageGpu.h` | Updated `RunBenchmark` signature to accept `HWND hwnd` |

## The Fix

Capture HWND on the UI thread, before the thread starts. Pass it by value into the lambda and forward it to `RunBenchmark`:

```cpp
// OnBnClickedBenchmark — UI thread
const HWND hwnd = GetSafeHwnd();          // safe: UI thread owns m_hWnd
m_benchThread = std::jthread([this, hwnd, filePath](const std::stop_token& stopToken) mutable
{
    SetThreadPriority(GetCurrentThread(), THREAD_PRIORITY_BELOW_NORMAL);

    // File I/O on worker thread — UI stays responsive
    std::vector<BYTE> testData = …ReadFile(filePath)…;

    if (testData.empty())
    {
        ::PostMessage(hwnd, WM_BENCH_DONE, 1 /*error*/, 0);  // uses captured HWND
        return;
    }

    RunBenchmark(stopToken, std::move(testData), hwnd);       // passes HWND through
});
```

```cpp
// RunBenchmark — now takes HWND as parameter, no m_hWnd access
void CPageGpu::RunBenchmark(…, const HWND hwnd)
{
    // hwnd is pre-captured — no GetSafeHwnd() call on worker thread
    for (…) {
        ::PostMessage(hwnd, WM_BENCH_RESULT, …);
    }
}
```

## Additional Improvements in This Patch

- **Worker thread priority**: `THREAD_PRIORITY_BELOW_NORMAL` so the benchmark doesn't starve other processes during measurement
- **Yield between measurements**: `SleepEx(5, FALSE)` between each algorithm so the thread scheduler stays responsive on single-core VM / low-core systems  
- **Error path in `OnBenchmarkDone`**: The `wParam` error flag is now checked and a proper error dialog is shown if the file could not be read (was previously silently ignored)


djdomi:
short update, gimme some hours i gonna fix the language situation you mention before :)

djdomi:
update the second: translate files should now be translated, since i only know german/english i can only confirm by this two, the rest has been automaticially translated.

djdomi:
Removed the unrelated `FinderNtfs.cpp` whitespace change — same stray diff as in #549, now cleaned up.

PR #561, opened 2026-06-15
Feature: support multiple folders in Individual Folder scan mode (#539)

### LINKED ISSUE #539: [Feature Request] Let "Indevidual Folder" support  binding multiple folders
Let "Indevidual Folder" support  binding multiple folders, such as 
        `c:\a\b\c; d:\e\f\g; h:\i\j\k; \\192.168.1.125\shared\adir`
This is similar to the binding of multiple drives that has already been implemented.

Image

djdomi:
> **Note:** I am not a programmer. This contribution was created with AI assistance. I have reviewed the changes for correctness to the best of my ability.

## Summary

- The **Individual Folder** scan mode now accepts **multiple folders** separated by semicolons (`;`)
- The **Browse** button appends to the existing selection instead of replacing it
- The **OK** button and validation logic handle all semicolon-separated paths correctly
- Updated the English label from `Individual &Folder` to `Individual &Folder(s)` to reflect the new capability

## How it works

The scan engine (`DirStatDoc.cpp`) already supports multiple paths via `SplitString()` — only the GUI needed updating:

1. **`OnOK()`** — splits the folder field by `;`, normalizes each path, and populates `m_drives` with all entries
2. **`UpdateButtons()`** — validates every path in the semicolon-separated list; OK is enabled only if all paths are valid
3. **`OnBnClickedBrowseButton()`** — appends the picked folder with `;` instead of overwriting
4. **`GetSelectedItems()`** — simplified to always return `m_drives` (populated for all modes after `OnOK()`)

## Test plan

- [ ] Enter two valid folders separated by `;` → scan starts and covers both trees
- [ ] Enter one valid and one invalid path → OK button stays disabled
- [ ] Click Browse twice → second folder is appended with `;`
- [ ] Single folder (existing behavior) still works as before

Closes #539


NoMoreFood:
Semi-colons are valid directory name characters.

harryytm:
I have tested the build of this PR and not working as expected, scan do not start and views remain empty.

As for divider, in my knowledge on WinDirStat code uses `|` as separated for drives.

windirstat/Dialogs/SelectDrivesDlg.cpp
```diff
@@ -699,11 +711,7 @@ BOOL CSelectDrivesDlg::PreTranslateMessage(MSG* pMsg)
 
 std::vector<std::wstring> CSelectDrivesDlg::GetSelectedItems() const
 {
-    if (m_radio == RADIO_TARGET_FOLDER)
-    {
-        return { m_folderName.GetString() };
-    }
-    return m_drives; // valid for both RADIO_TARGET_DRIVES_ALL and RADIO_TARGET_DRIVES_SUBSET
+    return m_drives; // valid for all modes after OnOK(); folder mode now populates m_drives too
 }
```
harryytm, line 715: I don't feel comfortable with this approach, m_drives should be remain its purpose for drives not paths.

djdomi: gimme some minutes I already fixing my suggestion entirely after some thinking of how to fix it useful. spare the time for the update :)

harryytm: If you have to implement this feature properly, you might need to make the command line work first and then pass that scan targets to the Finders engine, even me I still tried to understand how the scan engine architecture works, not quite a simple fix as you mentioned simply modifying the Select Drives dialog in the PR description.

Ideal syntax of the command line parameter should follow how Windows Shell handles it, example as follows:

```
WinDirStat.exe "C:\Windows" "C:\Program Files (x86)" "D:\Software Downloads"
```

But that being said, you need to study the existing code to know how it works, including the existing command-line syntax, or should consult the team (not me since I am not part of it) before making changes that too radical.

djdomi: Thanks for the advice. I also addressed your comments, assuming my understanding is correct.

I was thinking about whether it might be a better option to add a separate tab where the user can select folders through multiple choice. I am not sure if this is the intended direction for the project, though.

What I mean is something similar to the usual directory tree, but with checkboxes in front of each drive and folder. This would allow the user to select multiple drives or folders at once.
I have not implemented this yet, but I already considered it as a cleaner alternative to adding folders one by one. Even if the current pipe separated style is kept internally, a multi select interface would still make sense to me.

That said, I am not sure whether my first suggestion would be better, where each selected folder is still handled as a separate folder entry.

i meaned this dialog: 
image

Issue #107, opened 2024-10-22
Simplier Scans


srt54558:
Hello,
I'd like to request a new feature "fast scan" where the tree image is skipped for example. This one should speed up the scans (Good for HDDs and automation).

Also a hint for some folders like "this is a critical system directory" would be good to keep unexperienced users safe.

Tysm for keeping this app alive after all the years!!!!!!!!!!!!!!!!!!!!!

georgir:
I'm not the dev so grain of salt and all that, but I don't think "the tree image" is slowing down the scan at all. There's really nothing much that can be done to speed up a normal scan. There are some tricks with using last-modified dates on folders that could speed up a follow-up scan if you have the data from a previous one, but even that's more of a heuristic than anything, because it's only updated one level up on new/deleted files and not recursively or for all changes.

There may be things that can speed up the duplicate file scan, like only checksumming files that have potential duplicates based on file size, doing partial checksums or random offset comparisons and aborting the full checksumming if a difference is found and there are no more duplicate candidates for a file, prioritizing bigger files and ones that match by file name as well, completely ignoring files below a certain size threshold, etc. Don't know if these are already implemented.

NoMoreFood:
So the tree image will slow down the scan a little due to how some of the visual artifacts are updated.  It's not much though unless you happen to expand it to the level of where files are actively being adding.  

The duplicate check actually does some of what you saying.  First it doesn't bother doing checksums if there are no files with the same size.  After that it does partial checksum of 128 KiB.  It will only do full scans if there are multiple files with that same 128 KiB partial checksum.  I'm sure it could be improved further.

djdomi:
> **Transparency notice:** I am not a programmer. This contribution was developed with AI assistance (Claude Code). I reviewed the result, but the implementation was largely AI-generated.

I've submitted two pull requests that address this issue:

---

### PR #562 — CPU: float arithmetic + `par_unseq` parallel leaf shading

[feat/treemap-cpu-rendering](https://github.com/windirstat/windirstat/pull/562)

- Converts the inner `DrawCushion` loop from `double` to `float` arithmetic
- Renders all leaf rectangles in parallel via `std::execution::par_unseq`
- Optionally renders the bitmap on a background thread (`TreeMapAsyncRendering` setting) so the UI stays responsive during large scans

This PR alone could close #107 — it already gives a significant speedup on multi-core systems without any hardware requirement.

---

[...]

PR #562, opened 2026-06-15
treemap: float arithmetic, par_unseq leaf shading, optional async render

djdomi:
> **Transparency notice:** I am not a professional programmer. This PR was developed with the assistance of an AI tool (Claude). I reviewed and tested the changes to the best of my ability. Please review carefully.

## Summary

Addresses #107 (CPU rendering path). A separate PR will cover the GPU path.

**Core optimizations (always active):**
- `DrawCushion`: `double` → `float` throughout; `#pragma float_control(precise, off, push/pop)` scoped to the function enables FMA fusion and `rsqrtss`+Newton-Raphson for `1/sqrtf` under `/fp:fast`
- New job-list architecture: `BuildLayout` (UI thread) collects `LeafJob` structs (rect + surface + color), then `RenderLeafJobs` processes them with `std::execution::par_unseq` across the full leaf list — better parallelism than per-row
- `DrawOverlays` extracted as a separate step (BlitBitmap + folder frames + extension labels)
- `DrawTreeMap` is now a thin sync wrapper calling all three

**Thread-safety note:** `BuildLayout` intentionally stays on the UI thread because `TmiSetRectangle` writes to the shared `CItem` tree, which `FindItemByPoint` reads on mouse-move. `RenderLeafJobs` only touches pixel memory and is safe on a background thread.

**Optional async render (`COptions::TreeMapAsyncRendering`, default `false`):**
- Layout runs on the UI thread; pixel shading is handed off to `std::async`
- While rendering: solid black background + 3px blue progress strip at top
- Cancel flag (`std::atomic<bool>`) shared via `shared_ptr` with the lambda — `EmptyView` sets it and drains the future before clearing bitmaps
- Resize during async: stale result detected via stored `m_renderSize`, discarded, fresh render triggered
- Maintainer opt-in: set registry key `HKCU\...\TreeMap\TreeMapAsyncRendering=1`

**Future work (comment in source):** SIMD (SSE4/AVX2) inner loop for `DrawCushion` — pros/cons documented in code; recommended to profile auto-vectorization output first.

## Test plan

- [ ] Sync path (`TreeMapAsyncRendering=0`): scan a large directory, verify treemap renders correctly with correct colors and folder frames
- [ ] Async path (`TreeMapAsyncRendering=1`): verify blue progress bar appears, treemap completes and displays correctly
- [ ] Resize during async render: resize window while rendering — should trigger fresh render without crash
- [ ] Cancel: close/reopen directory while async render in progress — `EmptyView` should drain cleanly without hang
- [ ] Zoom + highlight: verify hit-testing (`FindItemByPoint`) and selection still work after async render
- [ ] No regressions in KDirStat and SequoiaView layout styles

:robot: Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/windirstat/Controls/TreeMap.cpp
+++ b/windirstat/Controls/TreeMap.cpp
@@ -252,27 +252,27 @@ void CTreeMap::RecurseCheckTree(const CItem* item)
 
 void CTreeMap::DrawTreeMap(CDC* pdc, CRect rc, CItem* root, const Options* options)
 {
-    // Validate parameters and options
+    auto layout = BuildLayout(pdc, rc, root, options);
+    if (layout.bitmapBits.empty()) return;
+    RenderLeafJobs(layout.bitmapBits, layout.leafJobs, nullptr, nullptr);
+    DrawOverlays(pdc, layout.renderArea, root, layout.bitmapBits, layout.folders);
+}
+
+CTreeMap::LayoutResult CTreeMap::BuildLayout(CDC* pdc, CRect rc, CItem* root, const Options* options)
+{
     ASSERT(pdc != nullptr && root != nullptr);
     if (pdc == nullptr || root == nullptr)
-    {
-        // Parameter check fallback
-        return;
-    }
+        return {};
 
 #ifdef _DEBUG
     RecurseCheckTree(root);
 #endif // _DEBUG
 
     if (options != nullptr)
-    {
         SetOptions(options);
-    }
 
     if (!PrepareRenderArea(pdc, rc, !m_options.grid))
-    {
-        return;
-    }
+        return {};
 
     CSelectStockObject soFont(pdc, DEFAULT_GUI_FONT);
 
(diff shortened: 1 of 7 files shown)


djdomi:
> **Transparency notice:** I am not a professional programmer. This comment was drafted with the assistance of an AI tool (Claude).

**Side note: Visual Studio toolset version**

While building this PR locally I noticed the project currently targets Platform Toolset `v143` (VS 2022). My local environment only has VS 2026 Build Tools (v145) installed, so I had to override the toolset flag to verify compilation.

The code changes in this PR are standard C++20 and should be compatible with any recent MSVC toolset (v143/v144/v145).

Just curious — are there any plans to raise the minimum Visual Studio requirement to VS 2022 or newer? Given that the project already uses C++20 features, v143 seems like a natural baseline. Happy to adjust the `.vcxproj` toolset entry if that aligns with the project direction.

NoMoreFood:
> Just curious — are there any plans to raise the minimum Visual Studio requirement to VS 2022 or newer? Given that the project already uses C++20 features, v143 seems like a natural baseline. Happy to adjust the `.vcxproj` toolset entry if that aligns with the project direction.

Not until Microsoft stops maintaining it. It's the last version that supports Windows 7 which, believe it or not, we still have some number of users using.

djdomi:

Pushed a follow-up fix: `e8ecf4e5`

**Bug:** `CSelectStockObject soFont(pdc, DEFAULT_GUI_FONT)` was placed inside the `if (m_options.showFolderFrames)` block in `DrawOverlays`. When folder frames are disabled but extension labels are enabled (`showExtensions = true`), `DrawTreeMapLabels` ran without `DEFAULT_GUI_FONT` selected — extension labels would render in whatever font the DC had at that point rather than the system GUI font.

In the original `DrawTreeMap`, the font was selected at the top of the function and stayed active through the entire rendering pass including `DrawTreeMapLabels`. The refactor accidentally narrowed the scope.

**Fix:** Moved `CSelectStockObject soFont` to the top of `DrawOverlays` so it covers both folder header rendering and `DrawTreeMapLabels`.

djdomi:

---

After submitting this PR I did an additional self-review and found three bugs that I've since fixed — documenting them here for transparency.

**Bug 1 — Progress bar invisible at render start** (`6e11c4a9`)
When an async render had just started and only a handful of leaves had been processed, the integer division `done * width / total` rounded down to zero and the blue progress bar was never drawn at all. The window just showed black with no feedback. Fixed with `std::max(5, ...)` so there is always at least a 5-pixel indicator as long as a render is running.

**Bug 2 — Double bitmap-select through fall-through in `OnDraw`** (`8cd135e3`)
The async-finalize path and the sync render path both selected `m_bitmap` into `dcmem`, then fell through to a second `CSelectObject sobmp2` at the bottom of the function, which selected it again immediately after the first RAII wrapper had de-selected it. Harmless in practice today, but a trap for future edits in between the two selects. Fixed by giving each path its own `BitBlt` + `DrawHighlights` + `return`, so the fall-through no longer exists.

**Bug 3 — Full window invalidated every 50 ms during render** (`8cd135e3`)
The 50 ms poll timer called `Invalidate(FALSE)` unconditionally, which caused a full `FillSolidRect` clear of the entire client area on every tick — even though the only thing that actually changes while rendering is the 3-pixel progress bar strip at the top. On a large window this is unnecessary GDI work. Fixed: `OnTimer` now calls `InvalidateRect` for only those 3 pixels while the render is still running, and falls back to full `Invalidate` once the future is ready.

Issue #465, opened 2026-03-22
WinDirStat does not start on Windows Server Core installations.


tmpforen-debug:
Since Server Core provides no graphical shell, I assume that the application requires a full GUI and does not offer an alternative interface. For comparison: TreeSize Free used to run on Server Core (before they dropped server support), presumably because they provided an alternative UI layer.

Adding support for Server Core would be a real improvement, as it would make the application usable in minimal‑footprint server environments.

NoMoreFood:
How would you expect it to work / look on server core?

tmpforen-debug:
> How would you expect it to work / look on server core?

From my perspective, a tree view showing folders along with their disk space usage would be perfectly fine and get the job done.

This is an example of TreeSize Free on a Server Core installation.
<img width="1282" height="391" alt="Image" src="https://github.com/user-attachments/assets/e785fad6-1ba4-4277-925e-abb5857ace54" />

harryytm:
> How would you expect it to work / look on server core?

It might be unrelated to this issue, but sometime I do use WinDirStat in Windows RE or PE as a File Explorer. XD

harryytm:
@tmpforen-debug looks like @NoMoreFood is testing something

NoMoreFood:
@tmpforen-debug 

Can you try this release? https://github.com/windirstat/windirstat/releases/tag/beta%2Fv2.5.8%2F2026-05-02

tmpforen-debug:
Great. I appreciate your effort and am thankful for it. It’s looking good and working well! Thank you.
<img width="682" height="487" alt="Image" src="https://github.com/user-attachments/assets/d0865864-47e3-48b4-9327-97f7904d3f6b" />

<img width="1745" height="941" alt="Image" src="https://github.com/user-attachments/assets/0811dc6d-8120-45c1-8407-9822343b045a" />

NoMoreFood:
Released with https://github.com/windirstat/windirstat/releases/tag/release/v2.6.0

PR #563, opened 2026-06-15
feat: GPU-accelerated treemap cushion rendering (D3D11 compute shader)

djdomi:

## Summary

- Adds an opt-in GPU rendering path for the cushion-shaded treemap bitmap
- New setting `TreeMapGpuRendering` (default: off) in the TreeMap options page
- When enabled and a D3D11 hardware device is available, all leaf-pixel Phong shading runs as a compute shader dispatch instead of the CPU `par_unseq` loop
- Falls back transparently to the CPU path on any D3D11 failure (no device, driver error, etc.)

## What this is NOT

**This PR is unrelated to PR #554** (which adds `GpuHasher.cpp` / `Shaders/GpuHash.hlsl` for SHA/MD5 file hashing used in duplicate detection).  
This PR renders the treemap bitmap on the GPU — a completely different subsystem.

## Dependency

**This PR depends on PR #562** (`feat/treemap-cpu-rendering`), which introduced the `LeafJob` / `BuildLayout` architecture that both the CPU `par_unseq` path and this GPU path rely on.  
Please review and merge PR #562 first. This PR is submitted as a draft until #562 lands.

## New files

| File | Purpose |
|------|---------|
| `windirstat/Shaders/TreeMapCushion.hlsl` | `cs_5_0` compute shader; one thread per leaf, iterates all pixels; NormalizeColor ported exactly from `CColorSpace` (integer arithmetic identical); `rsqrt()` matches CPU `/fp:fast` path |
| `windirstat/GpuRenderer.h` | Public API — `IsAvailable()` + `Render()` |
| `windirstat/GpuRenderer.cpp` | D3D11 singleton; hardware device only; `D3DCompile` at runtime with `D3DCOMPILE_OPTIMIZATION_LEVEL3`; amortised leaf buffer (grows ×1.5); exact-size bitmap buffer + CPU staging readback |

## Modified files

- `Controls/TreeMap.cpp` / `.h` — `RenderLeafJobs()` tries GPU first when cushion shading + option enabled
- `Options.h` / `.cpp` — `COptions::TreeMapGpuRendering` setting
- `resource.h` / `windirstat.rc` — `IDR_TREEMAPCUSHION_SHADER RCDATA` for the embedded HLSL
- `windirstat.vcxproj` — new files registered

## Test plan

- [ ] Build with VS 2022 (v143), x64 Debug and Release — no warnings/errors
- [ ] Enable `TreeMapGpuRendering`, scan a directory, verify visual output matches the CPU path
- [ ] Disable `TreeMapGpuRendering`, verify fallback to CPU path
- [ ] Run on a machine without a D3D11 hardware device — should silently fall back to CPU

### DIFF
--- a/windirstat/Controls/TreeMap.cpp
+++ b/windirstat/Controls/TreeMap.cpp
@@ -252,27 +252,27 @@ void CTreeMap::RecurseCheckTree(const CItem* item)
 
 void CTreeMap::DrawTreeMap(CDC* pdc, CRect rc, CItem* root, const Options* options)
 {
-    // Validate parameters and options
+    auto layout = BuildLayout(pdc, rc, root, options);
+    if (layout.bitmapBits.empty()) return;
+    RenderLeafJobs(layout.bitmapBits, layout.leafJobs, nullptr, nullptr);
+    DrawOverlays(pdc, layout.renderArea, root, layout.bitmapBits, layout.folders);
+}
+
+CTreeMap::LayoutResult CTreeMap::BuildLayout(CDC* pdc, CRect rc, CItem* root, const Options* options)
+{
     ASSERT(pdc != nullptr && root != nullptr);
     if (pdc == nullptr || root == nullptr)
-    {
-        // Parameter check fallback
-        return;
-    }
+        return {};
 
 #ifdef _DEBUG
     RecurseCheckTree(root);
 #endif // _DEBUG
 
     if (options != nullptr)
-    {
         SetOptions(options);
-    }
 
     if (!PrepareRenderArea(pdc, rc, !m_options.grid))
-    {
-        return;
-    }
+        return {};
 
     CSelectStockObject soFont(pdc, DEFAULT_GUI_FONT);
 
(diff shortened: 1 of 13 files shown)


djdomi:

---

After submitting this PR I did an additional self-review and found several bugs, all in the GPU-specific code paths. Documenting them here for transparency.

**Bug 1 — Cancel flag not checked before GPU dispatch** (`e10266ba`) ⚠️
When the user resized the window or started a new scan while an async GPU render was running, `EmptyView()` set `cancel = true` and blocked on `m_renderFuture.get()`. However, `RenderLeafJobsGpu` never checked the cancel flag — `GpuRenderer::Render` dispatched and waited for the GPU to finish regardless. This meant the UI thread could be blocked for the full render duration (potentially hundreds of milliseconds on large treemaps). Fixed by checking `cancel` before entering the GPU branch in `RenderLeafJobs`.

**Bug 2 — Progress bar invisible during GPU render** (`e10266ba`)
In GPU mode, `m_renderProgress` stays at 0 for the entire render duration and jumps to N in a single store once the GPU finishes. The timer fired ~20 times per second, each time computing `0 * width / N = 0` and drawing nothing. Fixed with the same `std::max(5, ...)` approach as the CPU branch (same fix is included in this PR via the rebase).

**Bug 3 — Device-lost not persisted after `Map()` failure** (`e10266ba`)
When `m_context->Map(m_bitmapStaging, ...)` failed (e.g. on device-lost), the code returned `false` so the CPU path correctly took over. But `m_available` was left as `true`, meaning every subsequent render would still dispatch the compute shader and copy the buffer before failing at `Map` again — doing real GPU work for nothing. Fixed by setting `m_available = false` on `Map` failure so the GPU path is disabled until the next app restart.

**Bug 4 — Silent init failures** (`e17f3bed`)
If `D3D11CreateDevice`, shader resource loading, or `D3DCompile` failed, `Initialize()` returned silently with no diagnostic output. The GPU option appeared to just do nothing, with no way to tell why from the outside. Added `OutputDebugStringW` at each failure point (plus the D3DCompile error blob text) so the reason is visible in a debugger or DebugView.

**Bug 5 — Double bitmap-select + full-window invalidation** (`e17f3bed`)
Same two issues as in the CPU branch (bugs 2 and 3 in the PR #562 comment) — fixed identically here since this PR shares `TreeMapView.cpp`.

harryytm:
Since WinDirStat is being used by system administrators, dependencies on additional libraries should be should be kept to a minimum, in order to maintain compatibility with Windows 7 or newer, as well as Windows Server Core environments that lack certain libraries, as noted in FR #465.

djdomi:
@nomorefood I'm sorry.i will resubmit to you ,codex killed my entire submissions in one go,soory if that makes some trouble

PR #473, opened 2026-03-27
✨ Quality: **Add optional side‑by‑side pane layout**

letho1608:
## Problem

**Severity**: `medium`
**File**: `**windirstat/WinDirStat.cpp** (or the source file that creates the main window/splitter – typically `MainFrame.cpp` or `WinDirStat.cpp` where the three panes are arranged)`

## Solution

Implemented the necessary changes to address the identified issue in `**windirstat/WinDirStat.cpp** (or the source file that creates the main window/splitter – typically `MainFrame.cpp` or `WinDirStat.cpp` where the three panes are arranged)`.

## Changes

- `windirstat/SideBySidePane.h` (new)
- `windirstat/SideBySidePane.cpp` (new)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

Closes #128

PR #566, opened 2026-06-16
feat: flexible 3-mode layout system for ultrawide monitors (fixes #128)

### LINKED ISSUE #128: [feature request] Alternative Layout
I'm happy to see a new version of WinDirStat — a big thank you for working on this!

Here's what version 2.0.3 looks like on my 49" monitor:

image

I would love to have the panes arranged side-by-side rather than stacked vertically.

Thanks in advance for considering this!

djdomi:

Picking up where PR #473 left off — this implements a full 3-mode layout system for issue #128.

**What this adds**

Three layout modes, selectable under Options → Layout (new tab):

**Mode 0 — Standard** *(default, nothing changes for existing users)*
The classic layout: file list and file types on top, treemap below. Untouched.

**Mode 1 — Side-by-side**
The treemap moves to the left or right side of the window. File list and file types share the opposite side in a vertical sub-splitter. A toggle in Options lets you choose which side the treemap sits on.

**Mode 2 — Wide (3 columns)**
All three panels — File List, Treemap, and File Types — are arranged in three independent columns. You freely assign which panel goes in which column via three dropdowns in Options. Designed for ultrawide monitors where vertical stacking wastes the most space.

**How the dividers behave**
- Dragging a divider all the way to the edge collapses that panel (existing drag-to-collapse behavior, extended to the 3-column case)
- A "Reset divider positions" button in Options → Layout restores all splitters to their defaults **immediately** — no restart required
- Switching layout modes requires a restart (uses the existing restart dialog)

**Implementation notes for maintainers**
- `COptions::LayoutMode` defaults to `0` (Standard). To make Side-by-side or Wide the out-of-box default, change the third argument in `Options.cpp` — there's a TODO comment marking the spot.
- The View menu has a commented-out `MENUITEM` that would open the Layout page directly (skipping the full Options dialog). Whether to expose this is left to maintainer discretion — there's a TODO comment in `windirstat.rc`.
- `CWdsSplitterWnd` was extended with a second splitter position (`m_splitterPos2`) to handle the 3-column case. The MFC member `m_nTrackColumn` is not accessible in the build toolchain used here (MSVC 14.51), so which divider was dragged is detected via delta comparison against stored positions.
- All 23 language files carry English fallback strings for the new UI elements.


NoMoreFood:
I don't like all the bloat from the localization (increases executable size).  Maybe you could have a procedural generated flyout menu that would display a lineart representation of the permutations for the user to select?

NoMoreFood:
I like this idea, but I'm going to work this change from here on out.  Thanks!

salvis:
Thank you for working on this, NoMoreFood and djdomi!

djdomi:
Your welcome, at least it seems i gave a good idea how it might be solved, i am happy about :)

PR #577, opened 2026-06-21
treemap: float arithmetic, par_unseq leaf shading, optional async render

djdomi:
> **Transparency notice:** I am not a professional programmer. This PR was developed with the assistance of an AI tool (Claude). I reviewed and tested the changes to the best of my ability. Please review carefully.

## Summary

Addresses #107 (CPU rendering path). A separate PR will cover the GPU path.

**Core optimizations (always active):**
- `DrawCushion`: `double` → `float` throughout; `#pragma float_control(precise, off, push/pop)` scoped to the function enables FMA fusion and `rsqrtss`+Newton-Raphson for `1/sqrtf` under `/fp:fast`
- New job-list architecture: `BuildLayout` (UI thread) collects `LeafJob` structs (rect + surface + color), then `RenderLeafJobs` processes them with `std::execution::par_unseq` across the full leaf list — better parallelism than per-row
- `DrawOverlays` extracted as a separate step (BlitBitmap + folder frames + extension labels)
- `DrawTreeMap` is now a thin sync wrapper calling all three

**Thread-safety note:** `BuildLayout` intentionally stays on the UI thread because `TmiSetRectangle` writes to the shared `CItem` tree, which `FindItemByPoint` reads on mouse-move. `RenderLeafJobs` only touches pixel memory and is safe on a background thread.

**Optional async render (`COptions::TreeMapAsyncRendering`, default `false`):**
- Layout runs on the UI thread; pixel shading is handed off to `std::async`
- While rendering: solid black background + 3px blue progress strip at top
- Cancel flag (`std::atomic<bool>`) shared via `shared_ptr` with the lambda — `EmptyView` sets it and drains the future before clearing bitmaps
- Resize during async: stale result detected via stored `m_renderSize`, discarded, fresh render triggered
- Maintainer opt-in: set registry key `HKCU\...\TreeMap\TreeMapAsyncRendering=1`

**Future work (comment in source):** SIMD (SSE4/AVX2) inner loop for `DrawCushion` — pros/cons documented in code; recommended to profile auto-vectorization output first.

## Test plan

- [ ] Sync path (`TreeMapAsyncRendering=0`): scan a large directory, verify treemap renders correctly with correct colors and folder frames
- [ ] Async path (`TreeMapAsyncRendering=1`): verify blue progress bar appears, treemap completes and displays correctly
- [ ] Resize during async render: resize window while rendering — should trigger fresh render without crash
- [ ] Cancel: close/reopen directory while async render in progress — `EmptyView` should drain cleanly without hang
- [ ] Zoom + highlight: verify hit-testing (`FindItemByPoint`) and selection still work after async render
- [ ] No regressions in KDirStat and SequoiaView layout styles

:robot: Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/windirstat/Controls/TreeMap.cpp
+++ b/windirstat/Controls/TreeMap.cpp
@@ -651,6 +651,10 @@ void CTreeMap::DrawOverlays(CDC* pdc, const CRect& rc, CItem* root,
     // Font must be active for both folder headers and DrawTreeMapLabels
     CSelectStockObject soFont(pdc, DEFAULT_GUI_FONT);
 
+    TEXTMETRIC tm{};
+    pdc->GetTextMetrics(&tm);
+    const int headerHeight = tm.tmHeight + 4;
+
     if (m_options.showFolderFrames)
     {
         CSetBkMode soBkMode(pdc, TRANSPARENT);
--- a/windirstat/Controls/TreeMap.cpp
+++ b/windirstat/Controls/TreeMap.cpp
@@ -350,27 +350,27 @@ void CTreeMap::RecurseCheckTree(const CItem* item)
 
 void CTreeMap::DrawTreeMap(CDC* pdc, CRect rc, CItem* root, const Options* options)
 {
-    // Validate parameters and options
+    auto layout = BuildLayout(pdc, rc, root, options);
+    if (layout.bitmapBits.empty()) return;
+    RenderLeafJobs(layout.bitmapBits, layout.leafJobs, nullptr, nullptr);
+    DrawOverlays(pdc, layout.renderArea, root, layout.bitmapBits, layout.folders);
+}
+
+CTreeMap::LayoutResult CTreeMap::BuildLayout(CDC* pdc, CRect rc, CItem* root, const Options* options)
+{
     ASSERT(pdc != nullptr && root != nullptr);
     if (pdc == nullptr || root == nullptr)
-    {
-        // Parameter check fallback
-        return;
-    }
+        return {};
 
 #ifdef _DEBUG
     RecurseCheckTree(root);
 #endif // _DEBUG
 
     if (options != nullptr)
-    {
         SetOptions(options);
-    }
 
     if (!PrepareRenderArea(pdc, rc, !m_options.grid))
-    {
-        return;
-    }
+        return {};
 
     CSelectStockObject soFont(pdc, DEFAULT_GUI_FONT);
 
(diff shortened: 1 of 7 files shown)


djdomi:

Hi, sorry for the re-opening — our fork (`wiki-mod/windirstat-fork`) was accidentally deleted by another AI tool (Codex), which closed all our PRs automatically. This PR is a direct continuation of #562 with identical content — no changes, just restored. The old PR's discussion history is preserved there for reference.

djdomi:
Updated to the latest version of sourcecode.

djdomi:
@NoMoreFood just a short question ,do you plan to take a look into here for a review?

 Also FYI, I schedule an update to this i with the next two days to fix the conflicts?

djdomi:

I invested some time today to bring this branch up to date against current master, and ran into something worth flagging in detail before I put more work into it, rather than quietly guessing at a direction and possibly building the wrong thing.

Since this branch was last rebased, it looks like there's been a fair amount of ongoing restructuring on the visualization side. `TreeMapView`/`TreeMap.cpp` and the other view files have changed substantially - a new shared `CGraphView` base class now owns window lifecycle, drawing, and hit-testing for TreeMap, Flame Graph, Storage Analytics, and the new Sunburst view, with each individual view only supplying its renderer-specific hooks (`RenderVisualization`, `DrawEmptyPlaceholder`, `CreateRenderBitmap`, and a few others). `OnDraw` itself is now implemented once in `CGraphView` and marked `final`, so individual views can no longer override it the way `CTreeMapView` used to.

That matters here specifically because of how this PR's optional async-rendering piece was built. The original problem it addresses is that on very large trees, computing the treemap layout and cushion-shading for every leaf can take long enough that the UI visibly freezes while it happens. The fix was to move that layout/shading work onto a background thread, with a small state machine (a pending-render flag, a cancellable atomic, a polling timer, and a finalize-on-ready step) living entirely inside `CTreeMapView::OnDraw`, so the view can show a progress indicator and keep painting while the background thread works, then blit the result once it's ready. Several of the follow-up commits on this branch (font-selection ordering, keeping the progress bar visible throughout, avoiding a redundant bitmap-select, a header-height fix) are all small corrections to that same mechanism.

With `OnDraw` now owned by `CGraphView` and shared across four views instead of belonging to `CTreeMapView` alone, I can't just drop that state machine back into place without picking one of two directions myself:

1. Build the async-render support into `CGraphView` itself, so any current or future view derived from it (Flame Graph, Storage Analytics, Sunburst, TreeMap) could opt into background rendering the same way, or
2. Find some narrower way to attach it to TreeMap specifically through the existing extension points, without touching the shared base - which would likely work against the intent of the refactor rather than fit into it.

The float-arithmetic and parallel-shading part of this PR is unrelated to any of this - it's internal to `TreeMap.cpp`'s per-leaf color math, and porting it onto the new `TreeMap.cpp` should be a much smaller, more mechanical job with comparatively little risk either way this gets decided.

Since you're the one who designed this shared view architecture, I'd rather ask directly than guess at what fits: do you already have a direction in mind for whether background/async rendering should become a `CGraphView`-level capability, or is that not something you want to take on right now at all? I'm happy to attempt either, but wanted to lay out the actual shape of the problem and get your read before sinking more time into a design that might not match where this is headed.

NoMoreFood:
I don't think there's enough value-added here to justify the code change so let's go ahead and close the PR.

PR #580, opened 2026-06-21
Sec-fix: quote path substitutions in User Defined Cleanup command lines

djdomi:

## Summary

`BuildUserDefinedCleanupCommandLine` substitutes the `%p`, `%n`, `%sp`, and `%sn` placeholders into user-supplied command templates without quoting the path values. The result is passed verbatim to `CMD.EXE /C`, which interprets unquoted `&`, `|`, `<`, `>`, and `^` as shell metacharacters.

Windows permits these characters in directory names. A directory named `legit project & calc.exe` produces this expansion for the template `ROBOCOPY %p D:\backup\`:

```
CMD.EXE /C ROBOCOPY legit project & calc.exe D:\backup\
```

CMD.EXE sees two separate commands: `ROBOCOPY legit project` and `calc.exe`.

## Where this matters

The scenario is most realistic on **shared network storage**:

1. A user has one or more UDCs configured (e.g. a backup or archive command).
2. The user scans an SMB share in WinDirStat.
3. An attacker with write access to the share has created a folder whose name contains CMD metacharacters (e.g. `reports & powershell -enc <payload>`).
4. The user right-clicks that folder in WinDirStat and applies a UDC — the payload executes in the user's security context.

The same applies to ZIP archives or any other source of attacker-controlled directory names.

**Limiting preconditions:** The user must have at least one UDC enabled (none are by default) and must explicitly apply it to the specific item.

## Fix

Replaces the four bare `ReplaceString` calls with a `SubstQuoted` lambda that:

- **If the placeholder is not quoted in the template** (e.g. bare `%p`): wraps the substituted value in double quotes → `"C:\path\to\item"`.
- **If the placeholder is already quoted** (e.g. `"%p"`): substitutes the raw value, avoiding double-quoting that would break paths with spaces.

Windows paths cannot contain `"`, so no inner-quote escaping is required.

**Side effect:** This also silently fixes the pre-existing issue where paths containing spaces would cause CMD to split the argument (`ROBOCOPY C:\My Documents D:\backup\` passes `C:\My` and `Documents` as separate arguments).

## Code quality note

I did a broader read-only review of the codebase while preparing this PR. The code is genuinely well-structured for a project of this age — consistent typed registry I/O via `Setting<T>`, clean NTFS→FinderBasic fallback, proper `regex_error` handling, and no widespread insecure patterns. That made it straightforward to audit.

I found five lower-severity observations I'm listing here **for maintainer awareness only**:

<details>
<summary>Informational findings (no action required)</summary>

**CR-01 — `CsvLoader.cpp`: `ITEMTYPE` cast without range check**
`static_cast<ITEMTYPE>(wcstoull(..., 16))` casts an external hex value to an enum with no validity check. A corrupted `.wds` file could supply an out-of-range value, resulting in UB in the dispatch code (in practice likely a silent wrong-type node, not a crash).

**CR-02 — `CsvLoader.cpp`: Incomplete JSON escape sequences**
The hand-rolled JSON parser handles only `\\` and `\"`. Missing: `\n`, `\r`, `\t`, `\b`, `\f`, `\uXXXX`. File names containing these sequences in a saved JSON scan result would round-trip back with literal `\n` etc. instead of the actual characters.

**CR-03 — `CsvLoader.cpp`: CSV parser doesn't implement RFC 4180 `""` quoting**
Embedded `"` characters in file names are written as `""` in quoted CSV fields (RFC 4180), but the current parser doesn't handle the `""` → `"` unescaping. Such entries would be mis-tokenized on reload.

**CR-04 — `CsvLoader.cpp`: `static z_stream*` in `GetCompressedResource` is not thread-safe**
The zlib decompression context is a static pointer reused across calls. No guard enforces the implicit assumption that the function is only called single-threaded at startup.

**CR-05 — `Property.cpp`: `std::stoi` without exception handling**
`Setting<std::vector<int>>::ReadPersistedProperty` calls `std::stoi(token)` without a `try/catch`. A trailing comma or non-numeric value in the registry entry would throw `std::invalid_argument` and crash the app at startup.

If any of these are worth addressing, feel free to let me know and I'll submit a separate PR.

</details>

---

Tested on Windows 11 x64. Verified both cases: bare `%p` and `"%p"` in the command template, with directory names containing `&`, `|`, and spaces.

### DIFF
--- a/windirstat/WinDirStatModel.cpp
+++ b/windirstat/WinDirStatModel.cpp
@@ -817,11 +817,28 @@ std::wstring CWinDirStatModel::BuildUserDefinedCleanupCommandLine(const std::wst
     ReplaceString(s, L"%sp", L">sp");
     ReplaceString(s, L"%sn", L">sn");
 
-    // Now substitute
-    ReplaceString(s, L">p", rootPath);
-    ReplaceString(s, L">n", rootName);
-    ReplaceString(s, L">sp", currentPath);
-    ReplaceString(s,L">sn", currentName);
+    // Substitute each marker, adding surrounding quotes unless the user's command
+    // template already wraps the placeholder in double quotes (e.g. "%p").
+    // Quoting prevents CMD metacharacter injection (&, |, ^, etc.) from path
+    // components. Windows paths cannot contain '"', so no inner-quote escaping
+    // is needed.
+    auto SubstQuoted = [&s](const std::wstring& marker, const std::wstring& value)
+    {
+        size_t pos = 0;
+        while ((pos = s.find(marker, pos)) != std::wstring::npos)
+        {
+            const bool alreadyQuoted = pos > 0 && s[pos - 1] == L'"' &&
+                pos + marker.size() < s.size() && s[pos + marker.size()] == L'"';
+            const std::wstring repl = alreadyQuoted ? value : (L'"' + value + L'"');
+            s.replace(pos, marker.size(), repl);
+            pos += repl.size();
+        }
+    };
+
+    SubstQuoted(L">p",  rootPath);
+    SubstQuoted(L">n",  rootName);
+    SubstQuoted(L">sp", currentPath);
+    SubstQuoted(L">sn", currentName);
 
     return s;
 }
--- a/windirstat/WinDirStatModel.cpp
+++ b/windirstat/WinDirStatModel.cpp
@@ -817,19 +817,34 @@ std::wstring CWinDirStatModel::BuildUserDefinedCleanupCommandLine(const std::wst
     ReplaceString(s, L"%sp", L">sp");
     ReplaceString(s, L"%sn", L">sn");
 
+    // A value substituted right before a literal '"' (ours or the user's own,
+    // see below) must not end in an odd run of backslashes: CommandLineToArgvW-style
+    // parsers treat a backslash immediately before a quote as escaping that quote
+    // rather than closing it, so e.g. a drive root "C:\" would swallow the closing
+    // quote and merge the rest of the command line into one argument. Doubling the
+    // trailing backslash run keeps the visible path unchanged while restoring the
+    // closing quote.
+    auto EscapeTrailingBackslashes = [](const std::wstring& value)
+    {
+        size_t count = 0;
+        while (count < value.size() && value[value.size() - 1 - count] == L'\\') ++count;
+        return value + std::wstring(count, L'\\');
+    };
+
     // Substitute each marker, adding surrounding quotes unless the user's command
     // template already wraps the placeholder in double quotes (e.g. "%p").
     // Quoting prevents CMD metacharacter injection (&, |, ^, etc.) from path
     // components. Windows paths cannot contain '"', so no inner-quote escaping
     // is needed.
-    auto SubstQuoted = [&s](const std::wstring& marker, const std::wstring& value)
+    auto SubstQuoted = [&s, &EscapeTrailingBackslashes](const std::wstring& marker, const std::wstring& value)
     {
         size_t pos = 0;
         while ((pos = s.find(marker, pos)) != std::wstring::npos)
         {
             const bool alreadyQuoted = pos > 0 && s[pos - 1] == L'"' &&
                 pos + marker.size() < s.size() && s[pos + marker.size()] == L'"';
-            const std::wstring repl = alreadyQuoted ? value : (L'"' + value + L'"');
+            const std::wstring escaped = EscapeTrailingBackslashes(value);
+            const std::wstring repl = alreadyQuoted ? escaped : (L'"' + escaped + L'"');
             s.replace(pos, marker.size(), repl);
             pos += repl.size();
         }


djdomi:

Hi, sorry for the re-opening — our fork (`wiki-mod/windirstat-fork`) was accidentally deleted by another AI tool (Codex), which closed all our PRs automatically. This PR is a direct continuation of #565 with identical content — no changes, just restored. The old PR's discussion history is preserved there for reference.

djdomi:
@NoMoreFood 

I opened a security-report on your  security site with the POC. even ifyou think its hypthetical it works very well, i tested it multiple times :D

NoMoreFood:
I haven't looked at this too hard yet. Do you think this is a breaking change?  Will it potentially affect folks who had manually added quotes?

NoMoreFood:
Actually, disregard my last question. I do see some issues with this.  There are cases were quoting may need to be done in special ways where do not want to add them automatically (escaping or maybe we want single quotes if double-embedded in PowerShell) .  Maybe we just do something like this:

    ReplaceString(s, L">qp", L'"' + rootPath + L"'");
    ReplaceString(s, L">qn", L'"' + rootName + L"'");
    ReplaceString(s, L">qsp", L'"' + currentPath + L"'");
    ReplaceString(s,L">qsn", L'"' + currentName + L"'");
    
    That way users can decide if we want it quoted or not....  Thoughts?

djdomi:
> Actually, disregard my last question. I do see some issues with this. There are cases were quoting may need to be done in special ways where do not want to add them automatically (escaping or maybe we want single quotes if double-embedded in PowerShell) . Maybe we just do something like this:
> 
> ```
> ReplaceString(s, L">qp", L'"' + rootPath + L"'");
> ReplaceString(s, L">qn", L'"' + rootName + L"'");
> ReplaceString(s, L">qsp", L'"' + currentPath + L"'");
> ReplaceString(s,L">qsn", L'"' + currentName + L"'");
> 
> That way users can decide if we want it quoted or not....  Thoughts?
> ```

Hei, gimme some time, i am preparing some tests and validations after your feedback, it might take a bit time on the run for the day. thanks in advance for the given.

windirstat/WinDirStatModel.cpp
```diff
@@ -817,11 +817,28 @@ std::wstring CWinDirStatModel::BuildUserDefinedCleanupCommandLine(const std::wst
...
+            const std::wstring repl = alreadyQuoted ? value : (L'"' + value + L'"');
+            s.replace(pos, marker.size(), repl);
+            pos += repl.size();
+        }
+    };
+
+    SubstQuoted(L">p",  rootPath);
+    SubstQuoted(L">n",  rootName);
+    SubstQuoted(L">sp", currentPath);
+    SubstQuoted(L">sn", currentName);
```
djdomi: i am not sure if my comment hit already, but pls ignore this for the moment until i propse the correct related update.

djdomi:

Thanks for looking at this closely, @NoMoreFood — you were right that the quoting needed more thought. I dug into it further and found a concrete case where the current fix breaks, though it's not quite the one you flagged.

## The case I missed

Any **drive root** (`C:\`, `D:\`, etc.) ends in a backslash. When `%p` expands to `C:\` and gets wrapped in quotes, the result is:

```
"C:\"
```

Standard Windows argument parsing (the `CommandLineToArgvW` convention almost every Win32 program uses, including `cmd.exe` itself for its own `/C` argument) treats a backslash immediately before a closing quote as *escaping* that quote, not closing it. So `"C:\"` never actually closes — the parser keeps consuming the rest of the command line as part of that one argument.

I verified this empirically with a minimal repro (~40 lines total: an 8-line program that prints its own `argv`, launched by a 30-line program that calls `CreateProcess("cmd.exe", "cmd.exe /C ...")` — the exact same pattern `BuildUserDefinedCleanupCommandLine` uses):

```
"C:\Users" second_argument   -> argv = ["C:\Users", "second_argument"]   (correct)
"C:\"      second_argument   -> argv = ["C:\" second_argument]           (broken - one argument, quote swallowed)
```

<details>
<summary>Repro code (build both with cl, run launchtest.exe with argvtest.exe in the same folder)</summary>

`argvtest.cpp` — just prints its own argv:

```cpp
#include <cstdio>
int main(int argc, char** argv)
{
    printf("argc=%d\n", argc);
    for (int i = 0; i < argc; ++i)
        printf("argv[%d] = [%s]\n", i, argv[i]);
    return 0;
}
```

`launchtest.cpp` — launches it the same way `CallUserDefinedCleanup` does (build once as-is for the broken case, once with `/DCONTROL_NO_TRAILING_BACKSLASH` for the control case):

```cpp
#include <windows.h>
#include <string>
#include <cstdio>

int main()
{
    // Mirrors WinDirStat's own pattern in BuildUserDefinedCleanupCommandLine /
    // CallUserDefinedCleanup:
    //   cmdline = "cmd.exe /C " + userCommandLine;
    //   CreateProcess(app.c_str(), cmdline.data(), ...);
#ifdef CONTROL_NO_TRAILING_BACKSLASH
    std::wstring cmdline = L"cmd.exe /C argvtest.exe \"C:\\Users\" second_argument";
#else
    std::wstring cmdline = L"cmd.exe /C argvtest.exe \"C:\\\" second_argument";
#endif
    wprintf(L"Launching: [%s]\n", cmdline.c_str());

    STARTUPINFOW si{ sizeof(si) };
    PROCESS_INFORMATION pi{};
    if (!CreateProcessW(L"C:\\Windows\\System32\\cmd.exe", cmdline.data(),
        nullptr, nullptr, FALSE, 0, nullptr, nullptr, &si, &pi))
    {
        printf("CreateProcess failed: %lu\n", GetLastError());
        return 1;
    }
    WaitForSingleObject(pi.hProcess, INFINITE);
    CloseHandle(pi.hThread);
    CloseHandle(pi.hProcess);
    return 0;
}
```

Build (from a VS Developer Command Prompt, MSVC v143 toolset, same as this project uses):

```
cl /nologo /EHsc /std:c++20 argvtest.cpp
cl /nologo /EHsc /std:c++20 launchtest.cpp /Fe:launchtest_bs.exe
cl /nologo /EHsc /std:c++20 /DCONTROL_NO_TRAILING_BACKSLASH launchtest.cpp /Fe:launchtest_control.exe
```

Run (both `argvtest.exe` and the two `launchtest_*.exe` need to be in the same folder):

```
.\launchtest_control.exe
  Launching: [cmd.exe /C argvtest.exe "C:\Users" second_argument]
  argc=3
  argv[0] = [...\argvtest.exe]
  argv[1] = [C:\Users]
  argv[2] = [second_argument]

.\launchtest_bs.exe
  Launching: [cmd.exe /C argvtest.exe "C:\" second_argument]
  argc=2
  argv[0] = [...\argvtest.exe]
  argv[1] = [C:" second_argument]
```

`launchtest_control.exe` splits correctly into two arguments; `launchtest_bs.exe` collapses into one, with the literal `"` from the command line ending up inside `argv[1]` instead of closing the quote — that's the swallowed-quote bug.

</details>

So as it stands, PR #580 would silently break **every** UDC action run against a scanned drive root — not a security bug, but a real regression, and a much more common trigger than the PowerShell-nesting case.

## On your `%qp`/`%qn` suggestion

Two things about that snippet specifically:

```cpp
ReplaceString(s, L">qp", L'"' + rootPath + L"'");
```

1. There's a mismatched quote in there — it opens with a double quote (`L'"'`) but closes with `L"'"`, which is a single quote/apostrophe, not a double quote. As written this would produce `"C:\Users\Test'` (opens `"`, closes `'`) — not valid quoting at all. I assume that's just a typo and you meant `L'"'` on both ends.
2. Even fixed, it would hit the exact same bug I describe above: unconditionally wrapping in quotes with no escaping means `%qp` on a drive root produces `"C:\"` and breaks the same way. So the trailing-backslash fix isn't something we can skip by moving to new placeholders — it would need to go into `%qp` too.

That's really my main objection to the opt-in-placeholder route: it doesn't avoid the quoting problem (both approaches need the same fix), it just relocates where the fix has to live — and it leaves the existing `%p`/`%n`/`%sp`/`%sn` (what's documented today, so what everyone's templates already use) unprotected unless people proactively migrate. Given there's already a working PoC (calc.exe pops from an unquoted `%p`), I don't think an opt-in mitigation actually closes the reported hole for existing users.

## Updated fix

Same approach as before (auto-quote unless the user already quoted it themselves), plus: before closing the quote, double any trailing backslash run on the value. This keeps the visible path identical (`C:\` still displays as `C:\`) but makes sure the quote actually closes:

```cpp
auto EscapeTrailingBackslashes = [](const std::wstring& value)
{
    size_t count = 0;
    while (count < value.size() && value[value.size() - 1 - count] == L'\\') ++count;
    return value + std::wstring(count, L'\\');
};
```

Applied to both branches (auto-quoted and already-quoted-by-the-user), since in both cases the value sits immediately before a literal `"`.

I've pushed this to the same branch. Re the PowerShell nested-quoting case you mentioned: I think that one's genuinely unsolvable in general at this layer (WinDirStat can't know if `%p` is landing inside a PowerShell single-quoted string, a here-string, or a plain cmd argument) — happy to add opt-in `%qp`-style placeholders as a *follow-up* for that specific advanced case if you still want them, but I'd suggest keeping the automatic quoting as the default fix so the reported vulnerability is actually closed for existing users.

If you do want the `%qp`/`%qn`/`%qsp`/`%qsn` placeholders as a follow-up, here's a corrected version of your snippet (fixed the mismatched quote, added the same trailing-backslash escaping) — not included in this PR, just leaving it here in case it's useful for a future one:

```cpp
ReplaceString(s, L">qp",  L'"' + EscapeTrailingBackslashes(rootPath)    + L'"');
ReplaceString(s, L">qn",  L'"' + EscapeTrailingBackslashes(rootName)    + L'"');
ReplaceString(s, L">qsp", L'"' + EscapeTrailingBackslashes(currentPath) + L'"');
ReplaceString(s, L">qsn", L'"' + EscapeTrailingBackslashes(currentName) + L'"');
```

NoMoreFood:
You are correct my example had a typo.  Here's a more practical example. If I wanted to do something like this:

`POWERSHELL -Command " '%p' | Set-Content '%p\go.txt' "`

By forcefully adding quotes, this command would no longer function as intended and could even introduce a new security problem due to parsing issues.  I think UDCs are a pretty advanced feature, and we probably need to just trust the user somewhat knows what they are doing.  That said, doing what I suggested (to encourage safe behaviors) or a best-effort regex-matched popup warning might be the best we could do without introducing a breaking change.  Or do nothing at all.  Just AI suggested it doesn't mean it's a good idea or a real problem ;-)

I'm open to further discussion though if you think you can safely address the concern.  Thank you for your contributions thus far.

djdomi:

Fair point, and a good example - I agree with your read on this one. And it's even a fair point to you that "AI suggested it" is definitely not the same as "it's correct," so I appreciate you actually poking holes in it instead of taking it at face value.

I don't think this is safely auto-fixable in general: WinDirStat has no way to know that `%p` is landing inside a PowerShell single-quoted string versus a plain cmd argument versus something else entirely, so I won't try to force-quote around that.

One small, low-risk option if you want it (happy to leave it out otherwise, not proposing it as part of this PR): right now our "already quoted, don't touch it" check only recognizes **double** quotes:

```cpp
const bool alreadyQuoted = pos > 0 && s[pos - 1] == L'"' &&
    pos + marker.size() < s.size() && s[pos + marker.size()] == L'"';
```

It could be extended to also recognize matching **single** quotes as "already quoted, leave it alone":

```cpp
const bool alreadyQuoted =
    pos > 0 && pos + marker.size() < s.size() &&
    ((s[pos - 1] == L'"'  && s[pos + marker.size()] == L'"') ||
     (s[pos - 1] == L'\'' && s[pos + marker.size()] == L'\''));
```

That would make your exact example above (`'%p'`) work unmodified, without changing behavior for the default (unquoted) case at all, since it only applies when the template author explicitly wrote matching quotes themselves.

To be clear about what it would and wouldn't cover: it wouldn't protect against a folder name containing an apostrophe breaking out of the single-quoted PowerShell string (Windows does allow `'` in file/folder names) - that's a separate, PowerShell-specific injection class that's arguably the "advanced feature, trust the user" territory you mentioned, since it requires the user to have opted into `'...'` quoting in a UDC template in the first place.

Also agree with your idea of a best-effort warning popup for risky-looking command lines (e.g. flagging a bare unquoted `%p`/`%n`/`%sp`/`%sn` when the UDC is saved) - that seems like a good complementary safety net regardless of what we do here, since it nudges people toward safer templates without changing behavior for anyone.

Let me know if you'd like this added, or if you'd rather leave this to documentation/user judgment as you suggested.

And thanks for taking the time to actually dig through this with me rather than just merging or dismissing it. I'm glad I could help close the simple, common case here, even though this whole thing started from a fairly hypothetical attack scenario. I don't think it's realistic to try to cover every nested-shell edge case like the one above - you could probably chase diminishing returns forever and end up with a pile of special-case handling for a handful of very unlikely setups. Drawing the line at "the common default is safe, advanced hand-crafted templates are on the user" seems like the right tradeoff to me.

As I mentioned when I opened this PR, WinDirStat is a genuinely well-maintained codebase, and beyond this one I haven't come across anything else that would count as a real, exploitable bug. So I'm fine with the approach as discussed above.

Let me know whether you'd like me to go ahead and add the single-quote handling, or whether you'd rather leave the fix as it is right now.

PR #598, opened 2026-07-01
FileWatcher Minor UX Tweak

harryytm:
- Change default width of Time column to show the timestamp in full out-of-the-box
- Change Time column sort sort in ascending order by default (older activity on top) to avoid scroll racing
- Configurable setting in Registry/INI for user to choose whether to retain the activity log or not on hiding FileWatcher view, enabled by default


NoMoreFood:
I don't think we need COptions::WatcherRetainLogOnHide.  Let's just remove the items if it's stopped/hidden.

harryytm:
> I don't think we need COptions::WatcherRetainLogOnHide. Let's just remove the items if it's stopped/hidden.

Certainly

harryytm:
> I don't think we need COptions::WatcherRetainLogOnHide. Let's just remove the items if it's stopped/hidden.

@NoMoreFood force-pushed the requested changes, anything else need to change as well?

PR #601, opened 2026-07-02
Fix Chinese language labels and neutral-script locale resolution

### LINKED ISSUE #600: Chinese language files mislabeled — lang_zh.txt shows as "Traditional / Taiwan" but contains Simplified Chinese text
Thank you for building and maintaining WinDirStat — it's a genuinely useful tool, and I appreciate that Chinese localization is already included.

**The bug**

In the language dropdown (Settings → General), the entry Windows displays as *"Chinese (Traditional, Taiwan) - 中文(台灣)"* is backed by `res/langs/lang_zh.txt`, but the file's actual content is entirely **Simplified** Chinese (简体中文), not Traditional. For example, line 1 reads `...目录统计信息...` — simplified glyphs.

Meanwhile the entry shown as *"Chinese (Traditional, Hong Kong SAR) - 中文(香港特別行政區)"* (`lang_zh-HK.txt`) does correctly contain Traditional Chinese (繁體中文) text.

**Root cause**

The label text isn't a translatable string in this repo — it's generated at runtime by Windows from the locale tag encoded in the filename (`Localization::GetLanguageList()` / `GetLocaleLanguage()` in `HelpersInterface.cpp`). The file uses the bare, "neutral" tag `zh`, which is an ambiguous/legacy BCP-47 tag whose resolved Windows display name isn't guaranteed to correctly convey "Simplified" (or even be stable across OS versions).

**Background: why Simplified and Traditional should be two distinct entries**

For historical reasons, Chinese (`zh`) splits into two major variants: Chinese Simplified (`zh-Hans` — the trailing `s` stands for "Simplified") and Chinese Traditional (`zh-Hant` — the trailing `t` stands for "Traditional"). Simplified and Traditional not only use different glyphs, but also diverge noticeably in IT/technical terminology, so treating them as two separate languages is the correct approach.

Regions that primarily use Simplified are Mainland China (`zh-CN`), Singapore (`zh-SG`), and Malaysia (`zh-MY`); regions that primarily use Traditional are Taiwan (`zh-TW`), Hong Kong (`zh-HK`), and Macau (`zh-MO`). Older software listed each `zh-<region>` combination individually, but enumerating six region-specific variants is unnecessarily granular. The modern web/software convention is to consolidate into two script-level tags instead: `zh-Hans` (Mainland-China-style wording) and `zh-Hant` (Taiwan-style wording, also broadly understood in Hong Kong/Macau).

**A second, related bug this surfaces**

Testing a fix locally turned up a second, more serious issue: even after picking the correctly-labeled Chinese entry and confirming the restart prompt, the UI silently stayed in English. Root cause: `Localization::LoadResource()` resolves the selected LANGID back to a lookup key via the legacy `GetLocaleInfo(LCID, LOCALE_SNAME)` API, which collapses "neutral script" locales to an arbitrary anchor region (e.g. a Simplified-neutral LANGID resolves back to `"zh-CN"`, a Traditional-neutral one to `"zh-TW"`) instead of round-tripping to the actual `zh-Hans`/`zh-Hant` tag used for the language file — so the string lookup never matches and the app quietly falls back to the English baseline it loads first. This is worth fixing in the same PR since a label-only fix would otherwise ship a still-broken language switch.

**Suggested fix**

1. Rename the two files to the standard BCP-47 script subtags instead of a bare or region-specific tag:
   - `lang_zh.txt` (Simplified content) → `lang_zh-Hans.txt`
   - `lang_zh-HK.txt` (Traditional content) → `lang_zh-Hant.txt`
2. Fix `Localization::LoadResource()`/`LoadExternalLanguage()` and `GetLocaleLanguage()` to resolve locale tags by **name** (`LCIDToLocaleName` with `LOCALE_ALLOW_NEUTRAL_NAMES`, `GetLocaleInfoEx`) instead of the legacy LCID-based `GetLocaleInfo`, so neutral-script locales resolve correctly both for string lookup and for the display label.

I verified locally (Win32 API calls + a full local build) that this produces working, content-accurate language switching with clean labels:
- `zh-Hans` → **"Chinese (Simplified) - 中文(简体)"**, loads the Simplified strings
- `zh-Hant` → **"Chinese (Traditional) - 中文(繁體)"**, loads the Traditional strings

I have a PR ready with this change.

DraftingDreamer:
Fixes #600

**Problem**

1. The Chinese-language entries in the language picker are mislabeled relative to their actual text content — `lang_zh.txt` (bare `zh` tag) contains Simplified Chinese but is shown as a Traditional/region-specific label; `lang_zh-HK.txt` correctly contains Traditional Chinese but is labeled specifically for Hong Kong even though the content isn't HK-specific.
2. Renaming the files alone isn't sufficient: `Localization::LoadResource()` (`Localization.cpp`) resolves the selected `LANGID` back into a lookup key via the legacy `GetLocaleInfo(LCID, LOCALE_SNAME)` API. That API collapses "neutral script" locales (like the ones we need for a clean Simplified/Traditional split) to an arbitrary anchor region — a Simplified-neutral LANGID resolves back to `"zh-CN"`, a Traditional-neutral one to `"zh-TW"` — instead of the actual `zh-Hans`/`zh-Hant` name used for the file. The string lookup then never matches, `LoadResource()` returns false, and the UI silently stays on the English baseline it loads first, even though a valid Chinese `LanguageId` was selected and persisted. Confirmed live with a local build: registry shows `LanguageId=0x7c04`, but the running app stayed entirely in English until this was fixed.

**Fix**

*Files (rename):*
- `lang_zh.txt` → `lang_zh-Hans.txt` (Simplified content)
- `lang_zh-HK.txt` → `lang_zh-Hant.txt` (Traditional content)

*Source (`Localization.cpp` / `Localization.h` / `HelpersInterface.cpp` / `HelpersInterface.h`):*
- `LoadResource()`: resolve the LANGID via `LCIDToLocaleName(..., LOCALE_ALLOW_NEUTRAL_NAMES)` (previously called with `flags=0`, and the correctly-resolved name was computed then discarded) and reuse that name directly for the BCP-47 resource/external-file lookup, instead of re-deriving it through the lossy `GetLocaleInfo(LCID, LOCALE_SNAME)` call.
- `LoadExternalLanguage()`: added a name-based overload so external override files also resolve by the correct neutral-script name.
- `GetLocaleLanguage()`: same fix for the dropdown label — resolves the display name via name-based `GetLocaleInfoEx` instead of the legacy LCID-based `GetLocaleInfo`, so the label reflects the neutral script (e.g. "Chinese (Traditional)") instead of collapsing to a region.

*Build/tooling, to keep everything consistent with the new filenames:*
- `windirstat.vcxproj` / `windirstat.vcxproj.filters`: updated the two `<Text Include>` entries.
- `windirstat/Build/Compress Strings.ps1`: the language-file discovery regex only accepted a 2-letter language code with an optional 2-letter *uppercase region* (`^lang_([a-z]{2}(?:-[A-Z]{2})?)\.txt$`), which would have silently dropped `lang_zh-Hans.txt`/`lang_zh-Hant.txt` from the build. Extended it to also accept a 4-letter Titlecase ISO-15924 script subtag.
- `setup/store/Package.appxmanifest.template`: declared Store resource languages `zh-HK`/`zh-CN` → `zh-Hant`/`zh-Hans`.
- `setup/store/language-screenshots.ps1`: updated the hardcoded `Code`/`LangId` table (`zh-hk`/3076, `zh`/30724 → `zh-hant`/31748, `zh-hans`/4).

*Intentionally not changed:* `MSI_CULTURE=zh-CN`/`zh-HK` inside the two files — those feed `setup/msi/make-multilingual.ps1`, which builds a WiX-localized MSI installer transform and needs a concrete regional culture. Unrelated to the in-app language picker.

## Test plan

- [x] Installed VS2022 Build Tools locally and did a full Release x64 build (0 warnings, 0 errors).
- [x] Ran the built app: language dropdown now shows "Chinese (Simplified) - 中文(简体)" and "Chinese (Traditional) - 中文(繁體)".
- [x] Selected each, restarted as prompted, and confirmed the entire UI actually switches — menus, dialogs, status bar, buttons all render correctly in the selected script, with locale-appropriate Traditional-Chinese (Taiwan-style) or Simplified-Chinese wording. Before the source fix, the UI silently stayed in English despite a valid `LanguageId` being selected and persisted.
- [x] Verified the updated `Compress Strings.ps1` regex matches both renamed files and all existing `lang_*.txt` files.


NoMoreFood:
@harryytm I defer to you on adjudication of this.  Let me know if it looks good to merge.

harryytm:
> @harryytm I defer to you on adjudication of this. Let me know if it looks good to merge.

Aside from renaming the lang file, the coding style violated OCP and DRY which I not quite like, and I really want to reproduce the problem to play with the bug, but I can't reproduce the bug.

PR #602, opened 2026-07-02
Alt to #601: keep lang_zh-HK.txt dedicated, add lang_zh-Hant.txt as a duplicate

DraftingDreamer:
Alternative to #601, opened for comparison per @harryytm's feedback on #600/#601 — **not meant to be merged alongside #601, pick one or the other.**

**What's different from #601**

`lang_zh-HK.txt` is left completely untouched here — still tagged `zh-HK` (LCID 3076), still labeled "Chinese (Traditional, Hong Kong SAR) - 中文(香港特別行政區)", still exclusively serving the Hong Kong SAR region as originally intended.

Instead of renaming it, this PR adds a new `lang_zh-Hant.txt` that is a byte-for-byte duplicate of `lang_zh-HK.txt`'s content, tagged with the generic Traditional-script `zh-Hant`. That way Traditional-Chinese Windows installs that aren't specifically configured for the Hong Kong region (Taiwan, Macau, or any other Traditional-Chinese locale) also get a working, correctly-labeled Traditional Chinese option — today they'd otherwise only see the Hong-Kong-specific entry or the mislabeled `zh` one.

Same as #601:
- `lang_zh.txt` → `lang_zh-Hans.txt` (fixes the core reported bug: Simplified content shown under a Traditional-looking label)
- The `Localization.cpp`/`HelpersInterface.cpp` locale-resolution fix (neutral-script LCIDs need to be resolved by name via `LCIDToLocaleName(..., LOCALE_ALLOW_NEUTRAL_NAMES)`/`GetLocaleInfoEx`, not the legacy LCID-based `GetLocaleInfo` which collapses them to an anchor region) — this part is required either way, it's what makes any `zh-Hans`/`zh-Hant` file actually load instead of silently falling back to English.
- Matching build/tooling updates (`.vcxproj`, `Compress Strings.ps1` regex, Store manifest, screenshot script).

**Trade-off vs #601**

- #601: one fewer file to keep in sync, but reroutes `zh-HK`'s dedicated tag to the generic script tag.
- This PR: keeps `zh-HK` exactly as-is (per @harryytm's request), at the cost of `lang_zh-Hant.txt` and `lang_zh-HK.txt` needing to be kept in sync by hand if the translation is ever updated (since they start out identical but are two separate files).

Happy to close whichever one isn't picked.


harryytm:
@DraftingDreamer PR #603 has been merged, which my take on PR #601 and #602 and partly inspired by your findings.

PR #603, opened 2026-07-03
Modernize GetLocaleString / Minor Refactoring

harryytm:
- replace deprecated GetLocaleInfo with GetLocaleInfoEx for primary local information retrieval
- minor refactoring and lines rearrangement

As seen in the screenshot zh now correctly named `Chinese - 中文` as I expected
image

Keep in mind this PR is only for GetLocaleString modernization, if any language files renaming needed, will leave it to another PR to maintain commit history readability.

### DIFF
--- a/windirstat/HelpersInterface.cpp
+++ b/windirstat/HelpersInterface.cpp
@@ -44,11 +44,21 @@ static std::wstring FormatLongLongNormal(ULONGLONG n)
 
 std::wstring GetLocaleString(const LCTYPE lctype, const LCID lcid)
 {
-    const int len = ::GetLocaleInfo(lcid, lctype, nullptr, 0);
+    wchar_t name[LOCALE_NAME_MAX_LENGTH]{};
+    if (LCIDToLocaleName(lcid, name, LOCALE_NAME_MAX_LENGTH, LOCALE_ALLOW_NEUTRAL_NAMES) == 0) return {};
+
+    return GetLocaleString(lctype, name);
+}
+
+std::wstring GetLocaleString(const LCTYPE lctype, const std::wstring& localeName)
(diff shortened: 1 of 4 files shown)


windirstat/HelpersInterface.cpp
```diff
@@ -44,11 +44,21 @@ static std::wstring FormatLongLongNormal(ULONGLONG n)
 
 std::wstring GetLocaleString(const LCTYPE lctype, const LCID lcid)
 {
-    const int len = ::GetLocaleInfo(lcid, lctype, nullptr, 0);
+    wchar_t name[LOCALE_NAME_MAX_LENGTH]{};
+    if (LCIDToLocaleName(lcid, name, LOCALE_NAME_MAX_LENGTH, LOCALE_ALLOW_NEUTRAL_NAMES) == 0) return {};
```
NoMoreFood, line 48: Would IsValidLocaleName() be a better fit here?

harryytm: I can add it, but this line not just for checking the localename, it also double as resolving locale name and write it to `name`, so even using IsValidLocaleName() it would become:

```cpp
    if (LCIDToLocaleName(lcid, name, LOCALE_NAME_MAX_LENGTH, LOCALE_ALLOW_NEUTRAL_NAMES) == 0 ||
        !IsValidLocaleName(name)) return {};
```

Which IsValidLocaleName(name) will very unlikely to be reach due to the short-circuit of == 0 being met if it is not a valid name, so it seems like not a good idea to have it here.

I can even try to add IsValidLocale() before it, but it just do the same as checking the return of LCIDToLocaleName().

Please let me know how you think about it.

Issue #593, opened 2026-07-01
remove empty folders


wincmd64:
This function needs to be expanded.
1) show a message how many were deleted, how many couldn't be deleted, and why.
2) also delete empty files (recursively).
3) ability to delete to recycle bin

djdomi:
I am not sure, where exactly you mean, but
<img width="352" height="274" alt="Image" src="https://github.com/user-attachments/assets/c1174513-350b-4258-acf6-c69318e9277a" />

Delete to recycle bin is already existing usually. would be great if you set your windirstat to English and provide some more Screenshots and explanation what you exactly mean

wincmd64:
> I am not sure, where exactly you mean

`Tools` -> `Remove empty folders` entry

djdomi:
@wincmd64 Okay, impressiv finding, i created a issue to not spam here on my fork to work on and will suggest an improvement :D was really shocked by testing it on an empty structure.

djdomi:
in addition, i see that there might be a miss-labeling

<img width="365" height="204" alt="Image" src="https://github.com/user-attachments/assets/aaf45a62-d7c1-49ff-bc52-14bf85a82eec" />
Empty Folder usually would understand a normal person to create one, and if not read correctly, even there might be pop-up, just smashing the button will empty the folder. will also take a look into it for a better wording :)


short updater later on:
@NoMoreFood Question, i think it might be Faster to change the line in the Code or Translations instead opening an extra PR for that.
Are you maybe intressted to rename ist to `Make Folder Empty (Delete all files/folders inside)` or doyou wihs aPR for it?

harryytm:
Technically not mislabeled because it means Empty (verb) Folder (noun), just as Windows `Empty Recycle Bin`.

However, if you really want to change it, I think that  `Empty The Folder` might be a better fit and is not too verbose as yours.

djdomi:
> However, if you really want to change it, I think that `Empty The Folder` might be a better fit and is not too verbose as yours.

Well yeh something in that direction i agree, as said i dont want go overboard but at least something bit more saying EY YO DANGER! :)

harryytm:
[...]

Issue #545, opened 2026-06-10
Feature request: show statistics for selected items


jernejs:
It would be nice if statusbar showed statistics (size and file count) when multiple items are selected in the folder tree:

<img width="2044" height="1016" alt="Image" src="https://github.com/user-attachments/assets/84c99e5e-77f4-4521-9a23-5760054c8209" />

Additionally, there should probably be a warning when clicking "Show the properties for selected items" and more than a few items are selected.

harryytm:
The `Physical Size: Σ` on the status bar is the total size of selected items since WinDirStat 1.x.

jernejs:
OK, not sure how I missed that. Still, also having a file count there would be nice.

harryytm:
> OK, not sure how I missed that. Still, also having a file count there would be nice.

Playing with the code try to add that, this is how the test build looks like, might need to add translations to the rest of the languages.

<img width="796" height="70" alt="Image" src="https://github.com/user-attachments/assets/935401b6-3397-4f75-8d37-f98c43bfaae8" />

harryytm:
> Additionally, there should probably be a warning when clicking "Show the properties for selected items" and more than a few items are selected.

As for this, I too not quite happy with how WDS handle show properties on multiple selected items, closing 100+ properties dialog is totally not fun at all, investigating how to change that too to align with how Windows handles that.

harryytm:
Have just found out @NoMoreFood already addressed this with fb7e6e96359205e77322f3d83c540bcb23ce1da2 which I quite happy with it, works seamlessly with [HashCheck 2.6.2.0](https://github.com/idrassi/HashCheck).

<img width="363" height="509" alt="Image" src="https://github.com/user-attachments/assets/6cb2e1a0-569a-4aa7-acf7-96bceba8093d" />

<img width="363" height="509" alt="Image" src="https://github.com/user-attachments/assets/accab886-f01d-4b6f-b6fb-97211f2e0947" />

> > Additionally, there should probably be a warning when clicking "Show the properties for selected items" and more than a few items are selected.
> 
> As for this, I too not quite happy with how WDS handle show properties on multiple selected items, closing 100+ properties dialog is totally not fun at all, investigating how to change that too to align with how Windows handles that.

harryytm:
My own UNOFFICIAL Development Preview Build

[WinDirStat.zip](https://github.com/user-attachments/files/28830865/WinDirStat.zip)

[...]

PR #604, opened 2026-07-03
Confirm before removing empty folders, default to Recycle Bin

djdomi:

Related to #593.

## Summary

`OnToolsRemoveEmpty` (Tools -> Remove empty folders) deletes every empty folder it finds with no confirmation and no Recycle Bin. This PR adds a two-step confirmation before anything is deleted, and makes the Recycle Bin the default outcome instead of permanent deletion.

## What This Actually Changes

Before this PR, clicking "Tools -> Remove empty folders" called `RemoveDirectory` directly on every folder whose entire subtree contains no files - immediately, permanently, with no dialog. If nothing was selected in the tree at the time, the code fell back to the whole scanned tree, so a single click could remove every empty folder on an entire scanned drive, including ones under system directories like `C:\Windows\`.

Now, before anything happens, a dialog lists every folder that would be affected and asks whether to move them to the Recycle Bin. Only if the user explicitly declines the Recycle Bin option does a second, separate dialog appear warning that permanent deletion cannot be undone.

For users, this does not remove any existing functionality - "Remove empty folders" still does the same detection and still removes the same folders when confirmed. The practical change is that nothing is deleted without the user first seeing what will be affected, and permanent deletion now requires an explicit extra step instead of being the only option.

## What This Fixes

This PR fixes the missing confirmation and missing Recycle Bin support that wiki-mod/windirstat-fork#5 raised as a safety gap. It does not change which folders are considered "empty" - that detection (a folder qualifies only if its entire subtree, including nested subfolders, contains zero files) is unchanged from before this PR.

## What Changed in Code

`WinDirStatModel.Actions.cpp`: `OnToolsRemoveEmpty` no longer calls `RemoveDirectory` or runs its own progress dialog. It builds the list of affected paths, shows a Yes/No/Cancel confirmation with that list (Yes = Recycle Bin, No = triggers the permanent-delete warning, Cancel = abort), and then delegates the actual deletion to the existing `DeletePhysicalItems`.

`WinDirStatModel.h` / `WinDirStatModel.cpp`: `DeletePhysicalItems` gets a new `skipWarning` parameter so it doesn't show its own separate confirmation on top of the two-step one `OnToolsRemoveEmpty` already displayed.

`res/LangStrings.h` and all 23 `res/langs/lang_*.txt`: two new strings, `IDS_REMOVE_EMPTY_TRASHs` and `IDS_REMOVE_EMPTY_PERMANENTs`, for the two new dialogs.

## Why This Matters for Users

A user who scans a whole drive and clicks "Remove empty folders" without selecting anything specific may not realize the action applies to everything, not just what's visible or focused in the tree. Without a preview or confirmation, that is a single irreversible click that can remove folders the user never intended to touch, including empty folders that other software or Windows itself expects to exist.

Making the Recycle Bin the default outcome means a mistake here is recoverable the same way a normal Explorer delete is, instead of being permanent by default.

## Scope

This PR only touches the confirmation/deletion-mechanism path of `OnToolsRemoveEmpty` and the shared `DeletePhysicalItems` helper. It does not change what counts as an "empty" folder, does not add the other two improvements requested in #593 (a post-operation summary, recursively removing empty files), and does not touch any other cleanup or deletion feature.

## Testing

Built and validated Debug/Release x64 locally; existing test suite passes. Have not yet manually clicked through the new dialogs in a running instance - will follow up with that before/alongside review.

## Edit

I missed a comment @harryytm left on #593 itself (not on this PR) on 2026-07-08, saying the current "Remove empty folders" feature is still experimental and that a proper implementation should have its own view showing the scan results before anything happens, pointing to [Remove-Empty-Directories](https://github.com/hxseven/Remove-Empty-Directories) by Jonas John as a reference design, instead of re-inventing that UX from scratch.

I was only watching this PR thread and did not see that comment until now. Given the Scope section above already says this PR does not add a post-operation summary or empty-file removal, does the dedicated-view suggestion also belong in a follow-up, or would you rather this PR grow to cover it?


harryytm:
Suggest make use of the checkbox in existing WDS CMessageBoxDlg API for user to select Recycle Bin or not, and use MB_YESNO, instead of using MB_YESNOCANCEL and no for permanent delete, that would be a more logical UX design for me.

Another suggestion is IDS_REMOVE_EMPTY_PERMANENTs and IDS_REMOVE_EMPTY_TRASHs might not be necessary, @NoMoreFood taken a lot of effort to remove bloat in the language file, see if it is possible to reuse existing IDS_CLEANUP_DELETE_BIN and IDS_CLEANUP_DELETE, and use code to replace the message when user check/uncheck the checkbox.

djdomi:

Thanks both for the review - I checked every point against the actual code and all four Codex findings turned out to be real, plus @harryytm's suggestions were both good improvements. Pushed a fix for all of it.

## Codex findings (all confirmed as real bugs)

- **Race condition**: `DeletePhysicalItems` deletes whatever currently exists under a path, not just what was empty at scan time - unlike the old direct `RemoveDirectory` call, which fails outright if the folder isn't actually empty anymore. Since time passes while the user reads the confirmation dialog(s), a folder could gain content in the meantime. Fixed by re-checking `PathIsDirectoryEmpty` on each folder right before deleting, immediately before the call to `DeletePhysicalItems`.
- **Recycle Bin offered where it doesn't exist**: confirmed - `ID_TOOLS_REMOVE_EMPTY` had no `hasRecycleBin`-style guard, so on a UNC/network path the "Recycle Bin" choice would have silently resulted in a permanent delete with no warning. Fixed by checking `IsLocalDrive` (the same check the existing "Delete to Recycle Bin" command already uses) before offering the option at all - if it's not available, the dialog goes straight to the permanent-delete warning instead.
- **Zoom left pointing at freed memory**: confirmed by tracing through `StartScanningEngine` - `IsAncestorOf` does include self, and if the zoomed item is one of the ones removed, `RemoveChild` frees it while zoom still points at it. Fixed at the root cause in `StartScanningEngine` rather than only in this handler, since the same logic path is shared by other refresh/delete callers: it now retargets zoom to the parent when the exact zoom item didn't survive a rescan.
- **Long paths**: confirmed this is real, but it's an existing limitation of `DeletePhysicalItems`'s shell-based item creation (`ILCreateFromPath` doesn't handle paths over `MAX_PATH`), already shared by every other feature that uses it. I didn't build a separate path just for this feature to work around a limitation that already exists everywhere else `DeletePhysicalItems` is used - flagging it here in case that's worth addressing separately at the shared-helper level.

## Harry's suggestions

Replaced the three-button Yes/No/Cancel (plus a conditional second dialog) with a single Yes/No dialog and a "Move to the Recycle Bin" checkbox, checked by default - matches the checkbox pattern already used elsewhere in `CMessageBoxDlg` (e.g. "do not show this again") instead of introducing a third button. Unchecking it and clicking Yes still falls through to the dedicated "this is permanent and cannot be undone" warning before anything happens, so that extra safety step for irreversible deletion is preserved.

On reusing `IDS_CLEANUP_DELETE_BIN`/`IDS_CLEANUP_DELETE`: I looked at these - they're formatted as tooltip-and-menu-label pairs (`"description\nShort Label"`), not full dialog sentences, so they don't fit the message body as-is. I did add one small new string for the checkbox label itself (`IDS_REMOVE_EMPTY_USE_TRASH`) since nothing existing fit that specific new interaction (a checkbox toggling Recycle Bin vs. permanent) - happy to revisit if there's a cleaner way to phrase this using existing strings that I'm missing.

harryytm:
> On reusing `IDS_CLEANUP_DELETE_BIN`/`IDS_CLEANUP_DELETE`: I looked at these - they're formatted as tooltip-and-menu-label pairs (`"description\nShort Label"`), not full dialog sentences, so they don't fit the message body as-is. I did add one small new string for the checkbox label itself (`IDS_REMOVE_EMPTY_USE_TRASH`) since nothing existing fit that specific new interaction (a checkbox toggling Recycle Bin vs. permanent) - happy to revisit if there's a cleaner way to phrase this using existing strings that I'm missing.

My apologies, the existing IDS should be IDS_DELETE_WARNING and IDS_EMPTY_FOLDER_WARNING used by the existing delete warning dialog as below:

```
IDS_DELETE_WARNING=IDS_DELETE_WARNING=You are about to delete the files listed below from your computer. It is your responsibility to verify whether deleting these files is safe. Deleting important system files could result in system malfunction.\n\nDo you wish to continue?
IDS_EMPTY_FOLDER_WARNING=Are you sure you want to delete all files in the following folders?
```

Seems that they might not fit for deleting empty folders, might be a good idea to create a new one, for example

```
IDS_REMOVE_EMPTY_FOLDER_WARNING=Are you sure you want to remove all the empty folders listed below?
```

I think one new IDS might be enough, just keep it simple.

djdomi:

Good call, thank you - down to a single new string now.

- Main dialog message: `IDS_REMOVE_EMPTY_FOLDER_WARNING` (your suggested wording, with a count added).
- Checkbox label: reused, not new - it's `IDS_CLEANUP_DELETE_BIN`'s existing menu-label half ("Delete (to Recycle Bin)"), split out the same way the menu itself already does.
- Permanent-delete confirmation: reused `IDS_DELETE_WARNING` as-is, the same one normal file deletion already shows.

Rebased onto master and pushed.

windirstat/WinDirStatModel.Actions.cpp
```diff
@@ -1422,40 +1428,67 @@ void CWinDirStatModel::OnToolsRemoveEmpty()
...
+
+    bool toTrashBin;
+    if (recycleBinAvailable)
     {
-        for (CItem* item : emptyDirs)
-        {
-            if (pdlg->IsCancelled()) break;
+        const auto result = CMessageBoxDlg::Show(
+            Localization::Format(IDS_REMOVE_EMPTY_FOLDER_WARNING, FormatCount(emptyDirs.size())), emptyPaths,
+            trashCheckboxText, true,
```
harryytm: Suggest use `IDS_MENU_DELETE_BIN=&Delete (to Recycle Bin)` directly instead, I don feel using `\n' is a reliable way to split the text, and adding additional code complexity just to extract the wanted string from IDS_CLEANUP_DELETE_BIN.

Test screenshot
Image

harryytm:
During testing, when selecting a directory like C:\ for `Remove Empty Folders`, there would be a UI freeze which feels like a crash but actually not, I think it might be a good idea to use the existing WDS API CProgressDlg wrapping the for-loop responsible for searching empty folders, to enable showing the user the search progress, and also allow the user to cancel the search.

djdomi:

Good catch, that makes sense - the search loop had no progress feedback or cancel at all before. Wrapped it in the same `CProgressDlg` pattern already used by `DeletePhysicalItems` for the deletion step, so it's cancellable now and shows progress instead of just an hourglass cursor.

Build was good, pushed. Harry, can you verify?

harryytm:
Hay you tested your build? The progress bar is not showing the progress as `Search Extension`.

djdomi:
> Hay you tested your build? The progress bar is not showing the progress as `Search Extension`.

Sorry, I am not getting the point yet - for me it works fine, showing a short bar which cycles and ends. I also searched the whole codebase and all language files for "Search Extension" and it does not appear anywhere in this project, so I am not sure what you are seeing.

Could you explain a bit more, or share a screenshot? Your last comment was a bit too short for me to follow up on, thanks in advance.

I put a freshly compiled test build of this exact branch here, in case there is a mismatch on your end: https://github.com/djdomi/windirstat-fork/releases/tag/test/safe-remove-empty-v1

harryytm:
> Sorry, I am not getting the point yet - for me it works fine, showing a short bar which cycles and ends. I also searched the whole codebase and all language files for "Search Extension" and it does not appear anywhere in this project, so I am not sure what you are seeing.

image

image

> I put a freshly compiled test build of this exact branch here, in case there is a mismatch on your end:

The build I was testing was ce19896d3d496abf020ee941d2890419b3ef66ad

> Wrapped it in the same `CProgressDlg` pattern already used by `DeletePhysicalItems` for the deletion step, so it's cancellable now and shows progress instead of just an hourglass cursor.

I haven't deep dive into your changes yet, but you can reference #462 which is a better case of `CProgressDlg` that most other parts of the code.

harryytm:
After a brief read of ce19896d3d496abf020ee941d2890419b3ef66ad , here is my initial findings:

1. You don't need to modify CProgressDlg, you only use it since the version in #462 already working, be aware AI don't follow Open-Closed Principle most of the time, they tend to make changes on things that are not broken and start breaking other parts of the code.
2. The changes in ce19896d3d496abf020ee941d2890419b3ef66ad used the bad use of CProgressDlg in existing code, which I think AI thought it was a good example of how CProgressDlg is used and the AI used it to generate the code.
3. CHANGELOG is not for us to update and it is the sovereignty of project maintainers, at lease this is my own rules.

djdomi:
Ill take alook into it, but due to my schedule ill be respondng with an pdate tue or wed next week, sorry for the bigger delay.

harryytm:
Found out what caused the UI freeze, created PR #606 to address that.

harryytm:
Update:
- I have submitted #607 which changed CMessageBoxDlg to show the progress of populating the list control, you can safely remove the CProgressDlg code in ce19896d3d496abf020ee941d2890419b3ef66ad since it not the source of the issue
- #606 have already merged and minimized the UI freeze impact, if #607 rejected, would ended up with 5 seconds of hour-glass wait, instead of a progress bar with cancel option
- `IDS_MENU_DELETE_BIN=&Delete (to Recycle Bin)` is more suitable for the checkbox, suggest use it and remove the complex code below
```cpp
    // The Recycle Bin menu command's label ("Delete (to Recycle Bin)") already says exactly
    // what checking this box will do, so reuse it instead of adding another string for it.
    const auto trashLabelParts = SplitString(Localization::Lookup(IDS_CLEANUP_DELETE_BIN), L'\n');
    const std::wstring& trashCheckboxText = trashLabelParts.back();
```
- please reference the follow existing code in windirstat\WinDirStat.cpp:394-406 and considered to simplify the CMessageBoxDlg calls pattern and make them cleaner and more readable in your PR
```cpp
    // Allow user to elevate if desired
    if (IsElevationAvailable() && COptions::ShowElevationPrompt && !hideApp)
    {
        if (const auto [nID, isChecked] =
            CMessageBoxDlg::Show(Localization::Lookup(IDS_ELEVATION_QUESTION), Localization::Lookup(IDS_DONT_SHOW_AGAIN),
                false, MB_YESNO | MB_ICONQUESTION, m_pMainWnd);
            (COptions::ShowElevationPrompt = !isChecked, nID == IDYES))
        {
            if (isChecked) COptions::AutoElevate = true;
            RunElevated(m_lpCmdLine);
            return FALSE;
        }
    }
```

harryytm:
030ad8c6aa7308ffcc00da0246b979c72c32c791 by @NoMoreFood have resolved the delay issue completely without the need of progress bar at all, I think it is better to let him commit the changes on his name rather that merging the PR, so I have withdrawal #607,  please wait for him to commit that changes.

harryytm:
@djdomi 
image
the wording in IDS_DELETE_WARNING not quite fit in this purpose, might be a good idea and necessary to create IDS_DELETE_EMPTY_FOLDERS_WARNING or resurrect IDS_REMOVE_EMPTY_PERMANENTs in your old commit in your PR.

```
IDS_REMOVE_EMPTY_PERMANENTs=This will permanently delete {} empty folder(s). This action cannot be undone.\n\nAre you sure you want to continue?
```

My take on the IDS_DELETE_EMPTY_FOLDERS_WARNING as follows to reduce code complexty.
```
IDS_DELETE_EMPTY_FOLDERS_WARNING=This will permanently delete all empty folders and cannot be undone.\n\nAre you sure you want to continue?
```

NoMoreFood:
To avoid localization bloat, I'm thinking let's just make do something like that we can reuse on many items.

<Menu Action Name>: The following items will be affected.  Do you wish to continue?

And then we could track the prompting with different variables.

harryytm:
> To avoid localization bloat, I'm thinking let's just make do something like that we can reuse on many items.
> 
> : The following items will be affected. Do you wish to continue?
> And then we could track the prompting with different variables.

I like and prefer the idea of strings that are universally reusable in many places instead of create a new one when adding new features.

djdomi:

@NoMoreFood that sounds like a genuinely good idea to me, a reusable template beats adding a new near-duplicate string every time a feature needs a confirmation dialog.

Before I touch anything: would you like me to build that reusable template as part of this PR (or a follow-up), or would you rather design and implement it yourself since it would likely touch other existing confirmation dialogs beyond just this one? Happy to do either, just do not want to guess at the shape of it and end up somewhere different from what you have in mind.

In the meantime I will hold off on any further string changes here so we do not need to redo work depending on your answer.

NoMoreFood:
You can take a crack at it.

harryytm:
@djdomi tested rebase latest head of main branch with 5dc599148cfee246c698ba977f7b7cf339f47d0f, the delay between selecting `Remove Empty Folder` and the warning window is non-existence, but the final remove stage after the warning message box does have a very long UI freeze, need to further investigate what caused the freeze.

djdomi:
Hi, I written a little test program I am testing since an hour.Some issues are related on how Windows reacts,but until I ended my tests I can't tell you more but 2-5ms on a 20k+ folder setup will see what happens

djdomi:

@harryytm I found the actual cause of the freeze after the warning dialog, and have real measurements to back up a proposed fix - writing this up in detail since it changes shared code used by more than just this feature, and I want to lay out exactly what was tested before picking a direction.

## Where the freeze is

It is in `CreateShellItemArray` (`windirstat/HelpersInterface.cpp`), used by `DeletePhysicalItems` to build the list of items handed to `IFileOperation` for the actual delete/recycle step:

```cpp
CComPtr<IShellItemArray> CreateShellItemArray(const std::vector<CItem*>& items)
{
    ...
    for (const auto& item : items)
    {
        PIDLIST_ABSOLUTE pidl = nullptr;
        if (item->IsTypeOrFlag(IT_MYCOMPUTER))
            SHGetKnownFolderIDList(FOLDERID_ComputerFolder, 0, nullptr, &pidl);
        else
            pidl = ILCreateFromPath(item->GetPath().c_str());   // <-- this, once per item
        ...
    }
    ...
}
```

`ILCreateFromPath` fully resolves each path through the shell namespace, one item at a time, with no way to batch it. That is fine for a normal manual multi-select, but "remove all empty folders" can easily produce thousands of items, and this loop runs serially before the delete even starts.

## What I tried and measured

I wrote a small standalone benchmark (not part of the project, just a throwaway console program using the same Windows APIs) so I would have real numbers instead of guessing:

- Created real, empty, disposable test folders (20,000 and separately 50,000 of them) on an SSD.
- Timed `ILCreateFromPath` against those paths, several trials, and reset between trials.
- Also timed wrapping the same loop in `std::execution::par` (parallelizing it), since that is the pattern already used elsewhere in this file for file deletion.
- Also timed the older, lighter-weight `SHSimpleIDListFromPath` as an alternative.

Results, averaged over repeated trials (raw logged output available if useful):

| Folders | Trials | `ILCreateFromPath` avg | `SHSimpleIDListFromPath` avg | Speedup |
|---|---|---|---|---|
| 20,000 | 10 | 62.7s | 4.8s | ~13x |
| 50,000 | 5 | 176.1s | 7.4s | ~24x |

Parallelizing `ILCreateFromPath` itself (`std::execution::par`) gave no real benefit (roughly 1x, sometimes slightly worse) - it looks like the shell call does not parallelize well, so that route is a dead end.

## The trade-off with `SHSimpleIDListFromPath`

It is faster because it does not verify the path actually exists - it just builds the ID list structurally. `ILCreateFromPath` binds through the real shell namespace and can fail (return null) for a path that is not there.

Since `CreateShellItemArray` is shared by three call sites (`DeletePhysicalItems`, `OnCleanupProperties`, and indirectly wherever else it gets reused), I did not want to just swap it in without checking whether losing that existence check could cause a real problem, so I also tested the actual deletion path end to end: built a shell item array from `SHSimpleIDListFromPath` results for 20 real disposable folders, ran the exact same `IFileOperation` recycle-bin delete the app uses, and confirmed all 20 disappeared from their original location and the Windows Recycle Bin item count increased by exactly 20. So the faster function still produces items `IFileOperation` handles correctly for an actual recycle-bin delete.

The only difference in practice: with `ILCreateFromPath`, an item that no longer exists gets silently skipped before it ever reaches `IFileOperation`. With `SHSimpleIDListFromPath`, it would be included and `IFileOperation` would have to deal with it not existing on its own (which it generally does gracefully for items it cannot find, since deletion operations are commonly used on selections that may have changed).

## What I would like your input on

Given the above, would you rather:
1. Swap `ILCreateFromPath` for `SHSimpleIDListFromPath` directly in `CreateShellItemArray`, so all three call sites benefit, accepting that a since-deleted item is no longer filtered out before reaching `IFileOperation`, or
2. Keep `ILCreateFromPath` as the default for the other call sites and only use the faster path specifically for the empty-folder deletion case (more code, but zero behavior change anywhere else), or
3. Something else you would prefer.

I lean towards option 1 since it is the smaller, more consistent change and the existence check does not seem load-bearing anywhere it is used, but wanted to lay out the actual numbers and the trade-off rather than just pick one.

harryytm:
@djdomi please study fb7e6e96359205e77322f3d83c540bcb23ce1da2 to investigate @NoMoreFood motives of writing the code in certain way first, or ask him directly.

that commit was to address the issue I raised
> Additionally, there should probably be a warning when clicking "Show the properties for selected items" and more than a few items are selected.
> 
> As for this, I too not quite happy with how WDS handle show properties on multiple selected items, closing 100+ properties dialog is totally not fun at all, investigating how to change that too to align with how Windows handles that. 

 _Originally posted by @harryytm in [#545](https://github.com/windirstat/windirstat/issues/545#issuecomment-4678203802)_

djdomi:

@harryytm, following up on the fb7e6e96 reference in your earlier comment, since I want to make sure I did not miss the actual concern.

CreateShellItemArray is also used by OnCleanupProperties. That path binds the resulting item array to BHID_DataObject and passes it to SHMultiFileProperties to open one combined Properties dialog.

If the binding fails, the code silently falls back to opening one Properties dialog per item. That is exactly the behavior fb7e6e96 was intended to eliminate. I assume this is the risk you were pointing out, and I agree that replacing ILCreateFromPath with SHSimpleIDListFromPath in shared code should not be allowed to reintroduce it unnoticed.

I created a small standalone test using five disposable folders. For both ILCreateFromPath and SHSimpleIDListFromPath:

* BindToHandler with BHID_DataObject succeeded.
* The resulting CF_HDROP contained all five file references.

My current reasoning is that this should also hold at larger scales. The relevant behavioral difference appears to be that ILCreateFromPath verifies that the path still exists, while SHSimpleIDListFromPath does not. Everything reaching OnCleanupProperties comes from items still present in the current scan, so that distinction should normally not affect this path.

However, I did not test two important parts:

* The actual SHMultiFileProperties call, because I deliberately avoided opening a real modal Properties dialog from an automated test.
* The multi-hundred or multi-thousand item case, which is the scale involved in the original problem.

Because of those gaps, I can only conclude that the change is probably safe. I cannot fully demonstrate that it preserves the behavior fb7e6e96 established.

Would you consider changing CreateShellItemArray safe for the Properties path as well, or would you prefer option 2, changing only the empty folder deletion path and leaving OnCleanupProperties untouched?

Separately, since this performance issue is pre-existing behavior in shared code that PR #604 merely exposed, rather than behavior introduced by #604, would you prefer that it be moved into a separate issue and PR instead of expanding this one further?

harryytm:
The major problem is I think the flow of Remove Empty Folders should be change to Search For Empty Folders, and then the search results would appear in the existing Search view, then allow user to select which folders to remove from there, instead of a one click solution.

That aside, your findings is valuable to improve the mass delete function, so I think a separate PR for optimize that would be more appropriate than being pat of this PR.

NoMoreFood:
Given the inactivity on this, close.  Will implement in another commit.

djdomi:
sorry, i forgot about this,  i am fine with it :)

PR #607, opened 2026-07-06
Message Box Show Cancelable Progress On Populating List

harryytm:
- Show progress of populating the list control and allow user to cancel
- Move population logic into a stand-alone member function for code readability

PR Integration
- Tested integration with #604


NoMoreFood:
How large does the list actually have to be to make this beneficial?

Try the alternative approach using ownerdata that I just committed to the PR.

harryytm:
I have tested your alternative approach using ownerdata and cut list population delay in #604 from 9 seconds wait in my progress bar approach to non-existance.

If you may you can close my PR and commit that change in your name.

harryytm:
I think it's better @NoMoreFood commit 030ad8c6aa7308ffcc00da0246b979c72c32c791 in his own name, withdrawing this PR.

harryytm:
@NoMoreFood 030ad8c6aa7308ffcc00da0246b979c72c32c791 have an issue that the checkbox text color becomes black in dark mode

image

Other than that your approach cut the delay time to non-existence which I think is the best solution.

NoMoreFood:
> have an issue that the checkbox text color becomes black in dark mode

Weird. I don't see that (and didn't change any code that should have affected that).  Is that in all dialog boxes or just that one?

harryytm:
Tried checkout 030ad8c6aa7308ffcc00da0246b979c72c32c791 and the changes in the commit affected other message box with checkbox enabled. My system is Windows 10 x64.
image
image

Gemini suggestions as follows does resolve the issue, but up to you how to resolve it, don't know if you have a better approach.

```cpp
HBRUSH CMessageBoxDlg::OnCtlColor(CDC* pDC, CWnd* pWnd, const UINT nCtlColor)
{
    // Set checkbox background to match dialog background in dark mode
    const int nID = pWnd->GetDlgCtrlID();
    if (nID == IDC_MESSAGE_CHECKBOX)
    {
        pDC->SetBkColor(DarkMode::WdsSysColor(COLOR_BTNFACE));
        if (DarkMode::IsDarkModeActive())                               // <-
            pDC->SetTextColor(DarkMode::WdsSysColor(COLOR_WINDOWTEXT)); // <-
        return m_checkboxBrush;
    }

    // Set icon and message text backgrounds to white in light mode
    if (!DarkMode::IsDarkModeActive() && (nID == IDC_MESSAGE_ICON || nID == IDC_MESSAGE_TEXT))
```

PR #615, opened 2026-07-08
Main Window Title Show Version

harryytm:


windirstat/WinDirStat.cpp
```diff
@@ -115,6 +116,12 @@ std::tuple<ULONGLONG, ULONGLONG> CDirStatApp::GetFreeDiskSpace(const std::wstrin
     return { u64total.QuadPart, u64free.QuadPart };
 }
 
+std::wstring CDirStatApp::GetAppTitle() const
+{
+    return std::format(L"{} {}{}.{}.{}", Localization::LookupNeutral(AFX_IDS_APP_TITLE),
```
NoMoreFood: Could we move this to the helpers for interface and make both this and CAboutDlg::GetAppVersion() utilize it?

PR #618, opened 2026-07-09
feat: flame graph view

### LINKED ISSUE #617: Feature Request: flame graph view
I would like to put forward a request for adding a flame graph view (and ideally making it the default one) alongside the treemap view. It might be a matter of personal taste, but in general I agree with the arguments in https://www.brendangregg.com/blog/2017-02-06/flamegraphs-vs-treemaps-vs-sunburst.html. I feel this can also provide a useful way of visualising the diffs from #486.

P403n1x87:
We add a flame graph view alongside the treemap view for an alternative visual representation of the file system hierarchical structure, with the size of each frame representing the physical size taken by each entry.

Closes #617


NoMoreFood:
@P403n1x87  Can you please allow edits to this PR?

P403n1x87:
@NoMoreFood it should be editable already

Screenshot 2026-07-10 at 15 44 57

NoMoreFood:
> @NoMoreFood it should be editable already

Yes, thanks.  Problem was on my end.  Can you test the changes I committed?  Functionally, I think the only thing that I didn't like was the single-click to drill down and the double-click to select.  Reversed it.

P403n1x87:
> > @NoMoreFood it should be editable already
> 
> Yes, thanks. Problem was on my end. Can you test the changes I committed? Functionally, I think the only thing that I didn't like was the single-click to drill down and the double-click to select. Reversed it.

@NoMoreFood cool thanks, will test as soon as I'm back on my Windows machine and report back. I didn't realise there was a double-click select event (mostly clanker-based work 😅 ) but that reversing makes sense 👍

P403n1x87:
Change looks good. I've been trying to improve the way thin frames render in the flame graph, but I'm not having a lot of success. When they get too thin, they look like they all merge.

image

PR #632, opened 2026-07-23
Minor Refactoring

harryytm:


harryytm:
@NoMoreFood Anything you don't like please let me know what to change or feel free to edit the PR directly.

NoMoreFood:
I've always followed the philosophy of defining the variables in close proximity to where they are actually used and there's not really added value to the lambda function.  Really the only thing I'd keep in here is using the wds::Mi/* wds::Ki that you added.

harryytm:
@NoMoreFood amended as requested in b44577e21a694300d21c65444b85c63110e16782

PR #692, opened 2026-09-14
Give IDC_FOLDER_HISTORY_COUNT its own control ID

djdomi:
> **Transparency notice:** I'm not the best programmer. This contribution might have been developed/improved with AI assistance. I reviewed and tested my result(s).

`IDC_FOLDER_HISTORY_COUNT` and `IDC_SEARCH_REGEX` are both defined as 1091.

Due to a review the numbers, i found a simple "bug" aka number doubling

It causes no trouble today, since control IDs are scoped to a dialog and those two live in different ones - `IDD_PAGE_ADVANCED` and `IDD_SEARCH`. So this is a trap for later rather than a present fault: if either control ever moved into the other dialog, it would break in a way that is hard to trace back to a line in `resource.h`.

`IDC_SEARCH_REGEX` keeps 1091, because 1091 to 1094 form the contiguous block of search controls. `IDC_FOLDER_HISTORY_COUNT` takes the next free value from `_APS_NEXT_CONTROL_VALUE` and moves to the end of the list, so the file stays in ascending order. Three lines in total.

I deliberately did not renumber anything else. Nothing in the code refers to these by number - I checked - so a wider renumbering would have changed well over a hundred lines without changing any behaviour, and would have been far harder to review than it is worth.

Checked every group in the file while I was in there - `ID_` (114 values), `IDC_` (147), `IDD_` (13), `IDR_` (5), `IDS_` (8). This was the only duplicate, and there are none left.

Verified with a Release x64 rebuild with warnings treated as errors. I have not run the program for this one, since the change resolves at build time.


NoMoreFood:
Thanks!

PR #696, opened 2026-09-17
Address Off-Viewport Modal Dialog Lockup

harryytm:
## Summary
Fixes an issue where modal dialogs (such as the Search dialog) render off-screen or vertically misaligned when the parent window is positioned near or beyond display edges, causing the main window to lock input.

This change restores monitor-aware boundary clamping within `CWnd::CenterWindow` in the custom UI framework and ensures dialogs center correctly after dynamic layout and height adjustments.

## Affected Versions
2.8.1 beta and later

## Root Cause
1. **Unconstrained Centering**: Window centering calculations computed parent-relative midpoints without clamping coordinates against the active monitor's work area (`MONITORINFO::rcWork`).
2. **Premature Centering**: In dialogs with dynamic control layouts (e.g., `SearchDlg`), centering was invoked before control expansion, causing post-init height adjustments to shift the window downward out of alignment.
3. **Over-Engineered Math & Type Mismatches**: Legacy centering logic relied on temporary `CSize` allocations, redundant 64-bit (`int64_t`) promotions, and `INT_MIN`/`INT_MAX` clamping, which introduced visual glitches and template type deduction errors (`E0304`) when pairing `int` variables with Win32 `LONG` bounds in `std::clamp`.

* **Streamlined Centering Logic**: Refactored `CWnd::CenterWindow` to calculate target origins using native 32-bit `LONG` coordinate arithmetic and custom `CRect` methods (`.Width()`, `.Height()`), removing redundant 64-bit integer promotions and temporary stack objects.
* **Monitor Work Area Clamping**: Added edge clamping using `std::clamp<LONG>` and `std::max<LONG>` against `MONITORINFO::rcWork` retrieved via `GetMonitorInfoW` and `MonitorFromWindow`.
* **Correct Initialization Timing**: Updated `SearchDlg` to pass `false` during `m_layout.OnInitDialog` and explicitly trigger `CenterWindow()` *after* window height adjustments complete.

## Steps to Reproduce (Before Fix)
1. Position the main window near or beyond the display edge.
2. Open a modal dialog (e.g., Search dialog).
3. Observe that the dialog renders partially or completely off-screen, locking input to the main window until closed via keyboard shortcut (`ESC`).


harryytm:
Refactoring in progress, set to draft

